Issue #8, opened 2021-10-15
operations/convertToGrey


lpatiny:
Is it ok to add a new method 'convertToGrey' that allows all the existing algorithms ?

I would also like to write as code:

`let greyImage = colorImage.grey({algorithm: 'minimum');`

https://github.com/image-js/image-js/blob/legacy/src/image/transform/grey.js

https://github.com/image-js/image-js/blob/legacy/src/image/transform/greyAlgorithms.js

targos:
That said, I think that for simplicity, `image.grey()` should be only applicable to RGB and RGBA images.

For generic color conversions in all directions, we keep `image.convertColor(newColorModel)`

opatiny:
Things left to do: make the grey test work and check they are no more bugs.

PR #20, opened 2021-11-12
Migrate grey algorithms

opatiny:
- feat: copy grey from image-js
- fix: fix all errors in greyAlgorithms
- docs: write grey algorithms docs
- fix: working on grey
- fix: fix grey clamp problem
- fix: fix all errors in grey
- test: fixing some of the grey test errors
- test: adapt invert test to new syntax
- test: add RGBA test for blur
- fix: invert had problem with alpha
- docs: add blur function doc
- feat: export copyAlpha from convertColors
- feat: add geometry methods to IJS
- fix: make options optional
- fix: keepAlpha option had a bugf
- fix: grey is working (#8)

### DIFF
--- a/src/operations/grey.ts
+++ b/src/operations/grey.ts
@@ -1,0 +1,93 @@
+import { clamp } from '../internal/clamp';
+import { getOutputImage } from '../internal/getOutputImage';
+import { GREY } from '../model/model';
+
+import { methods } from './greyAlgorithms';
+
+/**
+ * Call back that converts the RGB channels to grey. It will be clamped after.
+ *
+ * @callback GreyAlgorithmCallback
+ * @param {number} red - value of the red channel
+ * @param {number} green - value of the green channel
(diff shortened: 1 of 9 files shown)


src/operations/grey.ts
```diff
@@ -0,0 +1,107 @@
...
+    method = greyAlgorithms[algorithm];
+  }
+
+  let clamp = getClamp(newImage);
+
+  for (let i = 0; i < image.size; i++) {
+    const red = image.getValueByIndex(i, 0);
+    const green = image.getValueByIndex(i, 1);
+    const blue = image.getValueByIndex(i, 2);
+    const alpha = image.getValueByIndex(i, 3);
```
targos, line 91: please move this inside the `if(mergeAlpha)`. we don't want to get the alpha value unnecessarily

targos (changes requested):

PR #21, opened 2021-11-19
Migrate threshold algorithms

opatiny:
- feat: copy all thresholds from image-js
- foc: add computeThreshold tsdoc
- fix: migrating all thresholds to ts syntax
- fix: change histogram return type
- fix: make histogram a uint32 array
- feat: add all algorithms to threshold
- chore: remove useless comments
- test: add thresholds test
- test: fix histogram test

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "jpeg-js": "^0.4.3",
     "ml-convolution": "^2.0.0",
     "ml-matrix": "^6.8.0",
-    "num-sort": "^3.0.0",
     "tiff": "^5.0.2"
   }
 }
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "jpeg-js": "^0.4.3",
     "ml-convolution": "^2.0.0",
     "ml-matrix": "^6.8.0",
+    "num-sort": "^3.0.0",
```
targos, line 70:
```suggestion
```

seems unused

src/compute/histogram.ts
```diff
@@ -29,7 +29,7 @@ export function histogram(
     channel = 0;
   }
   validateChannel(channel, image);
-  const hist = new Array(image.maxValue + 1).fill(0);
+  const hist = new Uint32Array(image.maxValue + 1).fill(0);
```
targos, line 32:
```suggestion
  const hist = new Uint32Array(image.maxValue + 1);
```

PR #52, opened 2022-03-04
Migrate and debug `cannyEdge`

### LINKED ISSUE #39: Mask optimization
Currently we are converting in a clean way a boolean to a number using the method `boolToNumber`:

https://github.com/image-js/image-js-typescript/blob/d181906c1359157b2e87beb10b86ebf71929a125/src/Mask.ts#L185-L189

However javascript would allow to write directly (in which value is a number of boolean):

```
   const index = row * this.width + column; 
   this.data[index] = value && 1; 
```

but typescript would not be very happy.

The time difference is not negligible. In a small test the current approach is 40% slower than using the second approach. I think we should focus on speed in this library and add some ts-expect-error if needed.

opatiny:
- fix: add outer in erode and dilate
- feat: copy cannyEdge from image-js
- docs: enhance ts-docs
- fix: migrating cannyEdge to ts
- docs: add some ts-doc
- fix: enhancing cannyEdge
- feat: migrating hypotenuse
- fix (hypotenuse): remove eslint errors
- test: start hypotenuse test
- test (hypotenuse): test symmetry
- refactor: rename canny edge file
- docs (clamp): enhance tsdocs
- fix: compute convolution value can return raw
- feat: write rawDirectConvolution()
- chore: minor changes
- test (getConvolutionValue): test is passing
- test: rawDirectConvolution works
- fix: directConvolution works again
- feat: implement getIndex
- docs: add tsdocs
- fix: remove all cannyEdgeDetector errors
- refactor: rename guassian blur test
- docs: add some tsdocs
- test: add test for separable convolution
- fix(gaussianBlur): modify options
- test(gaussianBlur): test options were wrong
- fix(cannyEdge): code runs but result is wrong
- fix(cannyEdge): wrong variable used inside a loop
- fix(cannyEdge): function is finally working
- test(cannyEdge): working on function test
- chore: modified demo
- fix(cannyEdge): change return type
- fix(Mask): enhance setBit (close #39)
- fix(cannyEdge): finally seems to work properly
- test(cannyEdge): test getDirection()
- test(demo): move demos to subfunctions
- fix(cannyEdge): add hysteresis option
- test: refactor test images
- test(cannyEdge): created cannyEdge testcase
- test(demo): experimenting with demo
- chore: fix include errors
- test(demo): move test functions to own file


src/utils/clamp.ts
```diff
@@ -12,10 +18,22 @@ export function getClamp(image: IJS): ClampFunction {
...
 }
 
+/**
+ * Clamp value for 8-bit images.
+ *
+ * @param value - Value to clamp.
+ * @returns The clamped value.
+ */
 function clamp255(value: number): number {
   return Math.min(Math.max(value, 0), 255);
```
lpatiny, line 28:
```suggestion
  return value < 0 ? 0 : value > 255 ? 255 : value;
```

src/utils/clamp.ts
```diff
@@ -12,10 +18,22 @@ export function getClamp(image: IJS): ClampFunction {
...
 }
 
+/**
+ * Clamp value for 16-bit images.
+ *
+ * @param value - Value to clamp.
+ * @returns The clamped value.
+ */
 function clamp65535(value: number): number {
   return Math.min(Math.max(value, 0), 65535);
```
lpatiny, line 38:
```suggestion
  return value < 0 ? 0 : value > 65535 ? 65535 : value;
```

PR #70, opened 2022-04-01
Implement crop

opatiny:


src/operations/crop.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  checkProcessable(image, 'crop', {
+    bitDepth: [8, 16],
+  });
+
+  row = Math.round(row);
+  column = Math.round(column);
+  width = Math.round(width);
+  height = Math.round(height);
+
```
targos, line 59:
```suggestion
```

Let's remove the rounding. There's no obvious reason to support it.

opatiny: Done in 028e6cd0e5542c1b3025ca276e20e70686964cfe

PR #77, opened 2022-05-06
Implement `extract()`

opatiny:
- feat: implement extract
- docs: add extract demo
- docs: add testExtractRoi to demo

### DIFF
--- a/src/IJS.ts
+++ b/src/IJS.ts
@@ -503,7 +503,7 @@ export class IJS {
   }
 
   /**
-   * Extract the pixels of an image that as specified in a mask.
+   * Extract the pixels of an image, as specified in a mask.
    *
    * @param mask - The mask defining which pixels to keep.
    * @param options - Extract options.
--- a/src/IJS.ts
+++ b/src/IJS.ts
@@ -15,11 +15,10 @@ import {
   gradientFilter,
   derivativeFilter,
   DerivativeFilterOptions,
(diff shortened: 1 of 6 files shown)


src/IJS.ts
```diff
@@ -501,6 +502,17 @@ export class IJS {
     return crop(this, options);
   }
 
+  /**
+   * Extract the pixels of an image that as specified in a mask.
```
targos, line 506: This sentence doesn't sound right.

src/operations/extract.ts
```diff
@@ -0,0 +1,75 @@
+import { IJS } from '../IJS';
+import { Mask } from '../Mask';
+import { assert } from '../utils/assert';
+
+export interface ExtractOptions {
+  /**
+   * Specify from what row the image should be cropped.
```
targos, line 7: I would rather say that it's the row/column where the beginning of the mask is placed

src/operations/extract.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export function extract(
+  image: IJS,
+  mask: Mask,
+  options: ExtractOptions = {},
+): IJS {
+  const { row = 0, column = 0 } = options;
+
+  assert(
+    !(row >= image.height && column >= image.width),
```
targos, line 36: Left and top conditions are missing

Issue #56, opened 2022-03-11
Migrate `flipX()` and `flipY()` from `image-js`


opatiny:
These two functions seem pretty easy and could be migrated quickly. 

https://github.com/image-js/image-js/blob/8bfbcc524daafc35c35df62f5d18e7679deae7ee/src/image/filter/flipX.js

lpatiny:
We could implement

`flip(options={})`

options:
- axis: 'horizontal' (default) | 'vertical' | 'both'
- out

PR #80, opened 2022-05-12
feat: Flip

wadjih-bencheikh18:
closes : https://github.com/image-js/image-js-typescript/issues/56

### DIFF
--- a/src/filters/flip.ts
+++ b/src/filters/flip.ts
@@ -1,5 +1,4 @@
 import { IJS } from '../IJS';
-import { copyAlpha } from '../operations';
 import { getOutputImage } from '../utils/getOutputImage';
 
 import flipX from './flipX';
--- a/src/filters/flip.ts
+++ b/src/filters/flip.ts
@@ -1,6 +1,5 @@
 import { IJS } from '../IJS';
 import { copyAlpha } from '../operations';
-import { copyData } from '../utils/copyData';
 import { getOutputImage } from '../utils/getOutputImage';
 
 import flipX from './flipX';
@@ -29,8 +28,7 @@ export interface FlipOptions {
  */
 export function flip(image: IJS, options: FlipOptions = {}): IJS {
   const { axis = 'horizontal' } = options;
-  let newImage = getOutputImage(image, options);
-  copyData(image, newImage);
+  let newImage = getOutputImage(image, options, { clone: true });
   if (image.alpha) {
     copyAlpha(image, newImage);
   }
--- a/src/filters/flip.ts
+++ b/src/filters/flip.ts
@@ -1,0 +1,42 @@
+import { IJS } from '../IJS';
+import { copyAlpha } from '../operations';
+import { getOutputImage } from '../utils/getOutputImage';
+
+import flipX from './flipX';
+import flipY from './flipY';
+
+export interface FlipOptions {
+  /**
+   * Image to which the resulting image has to be put.
+   *
+   * @default 'horizontal'
+   */
+  axis?: 'horizontal' | 'vertical' | 'both';
+  /**
+   * Image to which the resulting image has to be put.
+   */
+  out?: IJS;
+}
+
+/**
+ *
+ * Apply a flip filter to an image.
+ *
+ * @param image - Image to process.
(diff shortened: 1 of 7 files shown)


src/filters/flipX.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export default function flipX(image: IJS): IJS {
+  checkProcessable(image, 'flipX', {
+    bitDepth: [8, 16],
+  });
+  for (let i = 0; i < image.height; i++) {
+    for (let j = 0; j < Math.floor(image.width / 2); j++) {
+      const colCurrent = j;
+      const colOpposite = image.width - j - 1;
+      const tmp = image.getPixel(i, colCurrent);
```
lpatiny, line 18: Please add a loop on the channels. getPixel would create an array that would have to be garbadcollected and this is time consuming.

We try in image-js to have correct performance

src/filters/flip.ts
```diff
@@ -0,0 +1,44 @@
...
+  copyData(image, newImage);
+  if (image.alpha) {
+    copyAlpha(image, newImage);
+  }
+  if (axis === 'horizontal') {
+    return flipX(newImage);
+  } else if (axis === 'vertical') {
+    return flipY(newImage);
+  } else {
+    return flipY(flipX(newImage));
```
targos, line 42: Can you add a couple tests for this case?

src/filters/flip.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param options - Flip options.
+ * @returns - The processed image.
+ */
+export function flip(image: IJS, options: FlipOptions = {}): IJS {
+  const { axis = 'horizontal' } = options;
+  let newImage = getOutputImage(image, options);
+  copyData(image, newImage);
+  if (image.alpha) {
+    copyAlpha(image, newImage);
+  }
```
targos, line 36: You can use the "clone: true" option of `getOutputImage` instead of this

src/filters/flipY.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns The processed image.
+ */
+export default function flipY(image: IJS): IJS {
+  checkProcessable(image, 'flipY', {
+    bitDepth: [8, 16],
+  });
+
+  for (let i = 0; i < Math.floor(image.height / 2); i++) {
+    for (let j = 0; j < image.width; j++) {
+      const ligneCurrent = i;
```
lpatiny, line 17: Please use currentRow and oppositeRow

src/filters/flipX.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param image - Image to process.
+ * @returns The processed image.
+ */
+export default function flipX(image: IJS): IJS {
+  checkProcessable(image, 'flipX', {
+    bitDepth: [8, 16],
+  });
+  for (let i = 0; i < image.height; i++) {
+    for (let j = 0; j < Math.floor(image.width / 2); j++) {
+      const colCurrent = j;
```
lpatiny, line 16: Please use currentCol and oppositeCol

lpatiny (commented):
Many thanks for this excellent contribution !

I did some small comments.

src/filters/flip.ts
wadjih-bencheikh18: Yeah no problem

src/filters/flipX.ts
wadjih-bencheikh18: fixed

src/filters/flip.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param image - Image to process.
+ * @param options - Flip options.
+ * @returns - The processed image.
+ */
+export function flip(image: IJS, options: FlipOptions = {}): IJS {
+  const { axis = 'horizontal' } = options;
+  let newImage = getOutputImage(image, options, { clone: true });
+  if (image.alpha) {
+    copyAlpha(image, newImage);
+  }
```
targos, line 34: I think alpha is also copied with `clone: true`

wadjih-bencheikh18: fixed thnx

PR #82, opened 2022-05-13
feat: PaintLine

wadjih-bencheikh18:


lpatiny:
I checked carefully and seems very well done, thanks !

Could you still add 2 test cases

Black image

0 0 0 0
0 0 0 0
0 0 0 0
0 0 0 0

And draw line from (0,1) to (3,2). I would expect something like:

0 0 0 0
1 1 0 0
0 0 1 1
0 0 0 0

And a second example

Black image

0 0 0 0
0 0 0 0
0 0 0 0
0 0 0 0

And draw line from (1,0) to (2,3). I would expect something like:

0 1 0 0
0 1 0 0
0 0 1 0
0 0 1 0

There is also a rebase problem because flip was now merged.

targos:
Maybe we should create a new subfolder for the paint-related functions? `src/paint` ?

src/operations/paintLine.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  column: number;
+}
+export interface PaintLineOptions {
+  /**
+   * Array of 3 elements (R, G, B),or 4 elements (R, G, B, A) default is red.
+   *
+   * @default [image.maxValue, 0, 0]
+   */
+  color?: [number, number, number] | [number, number, number, number];
```
targos, line 23:
```suggestion
  /**
   * Array of N elements (e.g. R, G, B or G, A), N being the number of channels.
   * Default is black.
   */
  color?: number[];
```

src/operations/paintLine.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Image to which the resulting image has to be put.
+   */
+  out?: IJS;
+}
+/**
+ * Paint a polyline defined by an array of points.
```
targos, line 30: This method doesn't paint a polyline.

src/operations/paintLine.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @returns The original painted image
+ */
+export function paintLine(
+  image: IJS,
+  from: Point,
+  to: Point,
+  options: PaintLineOptions = {},
+) {
+  let newImage = getOutputImage(image, options, { clone: true });
+  const { color = [newImage.maxValue, 0, 0] } = options;
```
targos, line 47: The default color must depend on the kind of image:
- Grey -> [0]
- GreyA -> [0, image.maxValue]
- RGB -> [0, 0, 0]
- etc.

You can create a function to generate it and do: `const { color = getDefaultColor(newImage) } = options;`

wadjih-bencheikh18: fixed

src/operations/paintLine.ts
wadjih-bencheikh18: yes sorry i also implemented polyline i will create the PR after mergin this one

src/operations/paintLine.ts
wadjih-bencheikh18: fixed

lpatiny (approved):

targos (changes requested):
I want to compare the behavior with the Canvas API before merging

src/operations/paintLine.ts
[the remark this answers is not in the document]
targos: You commented on an old version of the code.

src/utils/getDefaultColor.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @returns Black color
+ */
+export function getDefaultColor(image: IJS): number[] {
+  if (image.colorModel === ImageColorModel.GREY) return [0];
+  if (image.colorModel === ImageColorModel.GREYA) return [0, image.maxValue];
+  if (image.colorModel === ImageColorModel.RGB) return [0, 0, 0];
+  if (image.colorModel === ImageColorModel.RGBA) {
+    return [0, 0, 0, image.maxValue];
+  }
+  throw new Error(`image color model ${image.colorModel} is not compatible`);
```
stropitek, line 18: It'll be more readable if you convert this to a `switch(image.colorModel)`

wadjih-bencheikh18:
> Canvas API

would you please give me a link or explain how?

src/utils/getDefaultColor.ts
wadjih-bencheikh18: done

targos:
https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API

I'd like to compare the result of `paintLine` vs something like:

```js
ctx.moveTo(10, 10);
ctx.lineTo(20, 20);
ctx.stroke();
```

wadjih-bencheikh18:
there's a problem while testing with canvas api:
my function draw the first line but in canvas they use the seconde one 
Line_scan-conversion

wadjih-bencheikh18:
maybe we should implement [Xiaolin Wu's line algorithm](https://en.wikipedia.org/wiki/Xiaolin_Wu%27s_line_algorithm)

targos:
I see that the main part of the line is identical, so we can merge this.
I was going to open another issue about antialiasing.

targos:
Thank you for looking into it btw!

PR #83, opened 2022-05-15
feat: paintPolyline & polygon

wadjih-bencheikh18:

PR #84, opened 2022-05-15
feat: drawRectangle

wadjih-bencheikh18:


wadjih-bencheikh18:
does using `paintline`  make the function slower?

targos:
Currently yes, but it will be better when https://github.com/image-js/image-js-typescript/issues/89 is fixed.
In the future, I think we will need to implement more low-level and performant drawing functions that can be then used by drawLine, drawRectangle, etc.

wadjih-bencheikh18:
>In the future, I think we will need to implement more low-level and performant drawing functions that can be then used by drawLine, drawRectangle, etc.

Could you please assign to me this task

targos:
There are still some "paint" to change. See https://github.com/image-js/image-js-typescript/pull/84/files

wadjih-bencheikh18:
> There are still some "paint" to change. See https://github.com/image-js/image-js-typescript/pull/84/files

fixed, i saw some paint in draw line but it's already fixed in https://github.com/image-js/image-js-typescript/pull/83

src/draw/__tests__/drawRectangle.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    const out = new IJS(2, 3);
+    const image = testUtils.createRgbImage([
+      [100, 150, 200, 100, 150, 0],
+      [100, 200, 5, 3, 200, 0],
+      [150, 200, 255, 6, 150, 0],
+    ]);
+    const start = { row: 1, column: 0 };
+    const width = 3;
+    const height = 3;
+    const expected = image.drawRectangle(start, width, height, {
```
lpatiny, line 57: Please use only one object parameter that contains the following properties:
- row: default 0
- column: default 0
- width: default image.width
- height: default image.height
- color: default [0,0,0] (depends of the kind of image as you did)
- fill: default empty

lpatiny (changes requested):

src/draw/__tests__/drawRectangle.test.ts
wadjih-bencheikh18: If default color is black how could we create rectangle with fill and no color 
(i suggest adding color "none")

targos: `color: 'none'` SGTM

wadjih-bencheikh18: done

Issue #90, opened 2022-05-20
Implement low-level drawing functions


targos:
API TBD.

targos:
Use case: draw various primitive things (starting with lines) on an existing image without taking care of creating the output image and validating parameters.

wadjih-bencheikh18:
Is it preferable to create a new function or add an option param?

targos:
What do you have in mind?

wadjih-bencheikh18:
i think adding low boolean in options will be better

PR #97, opened 2022-05-22
feat: implement low-level drawLine

wadjih-bencheikh18:


src/draw/drawLine.ts
```diff
@@ -26,6 +26,10 @@ export interface DrawLineOptions {
    * Image to which the resulting image has to be put.
    */
   out?: IJS;
+  /**
+   * Enable low-level drawing.
+   */
+  low?: boolean;
```
targos, line 32: I'm not sure about this. The `low` option name isn't very self-explanatory and there is now a contradiction (if you pass `low: true` and `out`, ``out is ignored). 
@stropitek What do you think? This is to resolve https://github.com/image-js/image-js-typescript/issues/90

wadjih-bencheikh18: Maybe i can use type
`{Low:boolean}||{other params}`

stropitek: I don't know exactly how `getOutputImage` works.

I also don't understand what "low" means in this context. Would call it `clone` with a default value of false. Don't we have the option to create a new image in other methods as well @targos? API and default value should be consistent it that case

stropitek: Would say we can remove the `out` option. If image is cloned `out` image is cloned by the function.

targos: When I opened https://github.com/image-js/image-js-typescript/issues/90, I was thinking of making a private low-level API which would never take care or have the option to clone the original image. To me this is just about optimizing for the next methods that will draw a lot of lines with the same color/other options.

stropitek: ok got it, SGTM

maybe later we could expose those low-level method in a callback:
```js
const newImage = image.annotate((img) => {
  img.drawLine();
  img.drawCircle();
});
```

wadjih-bencheikh18: maybe we can implement a low-level drawline function(without any option params) and use it in original drawline

wadjih-bencheikh18:
@targos could i work on this ?
also would you please specifie how do you want me to impliment it ( I like daniel suggestion )

targos:
I think we need to design the API clearly before working on the code. I'll prepare a hackmd where we can do that collaboratively.

PR #100, opened 2022-05-24
feat: drawCircle

wadjih-bencheikh18:


targos:
Do you have a reference for the algorithm you used?

targos:
How does it compare to this canvas code for example: https://codesandbox.io/s/canvas-circle-forked-ot7pwh?file=/src/index.js

wadjih-bencheikh18:
> Do you have a reference for the algorithm you used?

No, i made it from scratch. do you want me to add comments ?

wadjih-bencheikh18:
> How does it compare to this canvas code for example: https://codesandbox.io/s/canvas-circle-86vutd?file=/src/index.js

Do you want me to add canvas test to compare the results?

lpatiny:
@wadjih-bencheikh18 In general making this kind of code from scratch may be quite dangerous (source of bug) and not optimized. If a code exists in a licence that is Apache, BSD or MIT we could get strongly inspired and cite them rather than reinventing it.

It is not a problem to reuse code if the licence allows it. In your case I have the impression that the code is not optimized and you could have a look at https://www.cs.helsinki.fi/group/goa/mallinnus/ympyrat/ymp1.html or some existing code.

wadjih-bencheikh18:
@lpatiny i finished implementing drawCircle

targos:
@wadjih-bencheikh18 It's nice that you take care of updating dependencies, but in the future please do it in a separate pull request (unless it's needed for the functionality).

Issue #105, opened 2022-06-16
Implement a function to correct color


opatiny:
Implement a function that allows to correct the colors in an image based on the known reference values of some pixels.

opatiny:
We will make a multivariate linear regression:
- x: a 20x12 matrix -> we have 12 variables for every square of the QP card (r, g, b, rg, gb, rb, r^2, ...)
- y: a 20x1 matrix -> the reference values for a channel (r, b or b)
We have to make one multivariate regression for each channel to get the correct model.

We use the `ml-regression-multivariate-linear` library.

PR #107, opened 2022-06-21
Implement `correctColor`

opatiny:
- test: colord library
- feat: add reference qp card data
- docs: enhance qp card comments
- refactor: rename correctColor
- test: test formatReferenceForMlr
- feat: getPixelByIndex
- feat: correctColor seems to work (#105)
- test: add qp card ref colors rgb
- refactor(correctColor): refactor code
- test(correctColor): algorithm is convincing (#105)
- fix(correctColor): handle RGBA images
- docs(demo): correctColor demo
- feat(jestMatchers): add error option
- feat(getImageColors): extract colors of an image
- test(correctColor): add test images
- fix(correctColor): normalise values for model
- test(correctColor): update snapshot


src/correctColor/__tests__/correctColor.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  it('offsets', () => {
+    const image = testUtils.load('correctColor/offsets.png');
+    const reference = testUtils.load('correctColor/test.png');
+
+    const measuredColors = getImageColors(image);
+    const referenceColors = getImageColors(reference);
+
+    const result = correctColor(image, measuredColors, referenceColors);
+
+    // all colors are nearly perfectly corrected, expect the gray
```
targos, line 139:
```suggestion
    // All colors are nearly perfectly corrected, except the gray.
```

test/jestMatchers.ts
```diff
@@ -17,51 +17,80 @@ interface MatcherResult {
...
+            const receivedValue = received.getValue(col, row, channel);
+            const expectedValue = expectedImage.getValue(col, row, channel);
+            if (Math.abs(receivedValue - expectedValue) > error) {
+              errorString = `Expected value at (${col}, ${row}) to be in range [${
+                expectedValue - error
+              },${expectedValue + error}], but got ${receivedValue}`;
+              break rowsLoop;
+            }
+          }
         }
```
targos, line 86: Do we need two different loops? I would have only done one with a default error tolerance of zero

opatiny: I think I left it like that so that the error message displays the whole pixel when the error is zero, but I can change it.

src/correctColor/__tests__/correctColor.test.ts
```diff
@@ -136,7 +136,7 @@ describe('correctColor', () => {
 
     const result = correctColor(image, measuredColors, referenceColors);
 
-    // all colors are nearly perfectly corrected, expect the gray
+    // All colors are nearly perfectly corrected, expect the gray.
```
targos, line 139: If "expect" is what you intended to write, I don't understand because the entire result is verified.

opatiny: The result is verified indeed. I just left this comment because I found it interesting that all the colors were corrected in the result ( no more offset in the squares), excepted in the grey square, where some variation is still visible.

targos: Then I think my suggestion was correct?
image

opatiny: Woopsie, my bad

Issue #66, opened 2022-04-01
Canny edge tests take a lot of time


targos:

opatiny:
I've used the following script to verify wether it is the function itself or the test environment that make it slow.
```js
const { readSync, ImageColorModel } = require('./lib/index.js');

console.time('a');
const image = readSync('./test/img/various/alphabet.jpg');
console.log({ width: image.width, height: image.height });
const grey = image.convertColor(ImageColorModel.GREY);
console.timeLog('a');
grey.cannyEdgeDetector();
console.timeLog('a');
```
I got the following results:
```bash
$ node testSpeed.js
{ width: 744, height: 800 }
a: 60.664ms
a: 212.597ms
```
We can see that the function execution takes about 0.14s, whereas in the tests, `cannyEdgeDetector` takes over a second on the same image. Here is the code that was run in the test:
```ts
console.time('a');

    const image = testUtils
      .load('various/alphabet.jpg')
      .convertColor(ImageColorModel.GREY);
    console.timeLog('a');

    console.log({ width: image.width, height: image.height });

    const expected = testUtils
      .load('morphology/alphabetCannyEdge.png')
      .threshold();
    console.timeLog('a');

    const result = image.cannyEdgeDetector();
    console.timeLog('a');
    expect(result).toMatchMask(expected);

    console.timeEnd('a');
````

And the results:
```bash
      a: 216 ms
      { width: 744, height: 800 }
      a: 280 ms
      a: 2096 ms
      a: 2442 ms
```

Conclusions: I don't really know how to make the tests faster. Any clues @targos ?

Issue #26, opened 2021-12-02
Complete code coverage


targos:
This a tracking issue. It does not have to be fixed in one go.

https://codecov.io/gh/image-js/image-js-typescript/branch/main

opatiny:
They are a few functions for which the coverage should really be enhanced:
- `transform`
- `rotate`: has two skipped tests right now, comparing results with opencv
- `readImg`: has no test at all currently

PR #116, opened 2022-06-28
Implement `getConvexHullMask`

opatiny:
- feat: add kind option on getMask
- chore: copy mbr from image-js
- fix(getMask): handle default options
- test(getContourMask): test various options
- docs(demo): getContourMask demo
- fix(getContourMask): default options were false
- test: enhance coverage of IJS
- test: enhance Mask coverage
- test(resize): working on resize test
- refactor: rename solidFill test file
- test(resize): enhancing coverage
- test(convertColor): enhance coverage
- test(transform): enhance coverage
- test(cannyEdgeDetector): working on #66
- test(getContouMask): fix allowCorners test
- test(rotate): enhance coverage
- docs(demo): add rotate demo
- test(decode): enhance coverage
- test(open): fix coverage
- test(convertDepth): fix coverage
- test(copyTo): fix coverage
- test(greyAlgorithms): enhance coverage
- test(merge): add missing tests
- test(greyAlgorithms): enhance coverage
- test: continue to enhance coverage (#26)
- chore: modify gitignore
- fix(copyTo): default options
- feat: copy monotoneChainConvexHull
- refactor: move Point to utils
- fix(mcch): use object point instead of array
- fix(getBorderPoints): use object Point
- refactor(getMask): move interfaces to subfunctions
- fix: Mask cutom inspect
- feat: implement getConvexHullMask
- fix(getConvexHullMask): handle 1 pixel ROI
- test(getConvexHullMask): add function test
- test(drawPolygon): add skipped tests showing bug
- feat(arrayPointsToObjects): points conversion tool
- fix(drawPolygon): filled option had a bug
- feat: implement getConvexHullMask


src/draw/drawPolygonOnMask.ts
```diff
@@ -34,31 +36,12 @@ export function drawPolygonOnMask(
...
-      }
-    }
-  }
-
-  for (let row = 0; row < mask.height; row++) {
-    for (let column = 0; column < mask.width; column++) {
-      if (polygonMask.getBit(column, row)) {
+  for (let row = 0; row < newMask.height; row++) {
+    for (let column = 0; column < newMask.width; column++) {
+      if (robustPointInPolygon(arrayPoints, [column, row]) === -1) {
```
lpatiny, line 44: Could you store the state of the previous pixel (on or off) and if the next pixel has the same state, robustPointInPolygon will yield the same result. This is a low hanging fruit that should improve the speed quite a lot (likely many order of magnitude) before implementing a better algorithm.

opatiny: Yes, I'll do that

opatiny: Actually I don't think this could work, because I don't have the "next pixel state". I am drawing a polygon onto an existing mask and the only data I have about the polygon are it's vertices.

Did I misunderstand what you meant?

lpatiny: Ok I see you don't draw first the polygon before filling it. So we will improve this later.

src/draw/__tests__/drawPolygonOnIjs.test.ts
```diff
@@ -129,7 +130,7 @@ describe('drawPolygon on IJS', () => {
     expect(result).not.toBe(image);
   });
   // the two following tests fail because there is a bug with filled polygons
```
targos: this comment can be removed now

opatiny: I had already done it in the latest commits.

Why is the comment still there when I click on the code snipped above, but not there when I go under the "Files changed" tab of the PR?

src/draw/drawPolygonOnMask.ts
opatiny: Ok I understand what you meant now, I can try to draw the polyline first and then check if the value stays the same.

opatiny: Yes, I'd rather already merge this PR, which is already really big and do that afterwards.

opatiny: I made an issue so that we don't forget to do this: #119

src/draw/__tests__/drawPolygonOnIjs.test.ts
targos: Not sure, maybe I was looking at an outdated diff.

src/geometry/__tests__/transform.test.ts
```diff
@@ -7,11 +7,40 @@ describe('transform with a transformation matrix', () => {
...
+    const transformed = img.transform(translation, {
       width: 16,
       height: 20,
     });
+    writeSync('src/geometry/__tests__/transformed.png', transformed);
 
     expect(transformed).toMatchImage('opencv/testTranslation.png');
   });
+
+  // is this the expected result for the fullImage option??
```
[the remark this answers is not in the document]
opatiny: I am not entirely sure what is the desired behaviour for this option, hence the skipped test.

PR #123, opened 2022-07-01
feat: impliment Bresenham's algorithm in drawLine

wadjih-bencheikh18:


targos:
Please install and use https://github.com/w8r/bresenham-zingl (don't copy the code)

wadjih-bencheikh18:
> Please install and use https://github.com/w8r/bresenham-zingl (don't copy the code)

bresenham-zingl does not have ts version, it may cause issues

wadjih-bencheikh18:
Do i have to create new drawline using Bresenham's algorithm or replace the function that already exist?

targos:
The goal is to replace the implementation, in other words call the library from `drawLineOnIjs` and `drawLineOnMask`

targos:
> > Please install and use https://github.com/w8r/bresenham-zingl (don't copy the code)
> 
> bresenham-zingl does not have ts version, it may cause issues

You can create a file in `src/types/bresenham-zingl.d.ts` and define the types that we need.

wadjih-bencheikh18:
> The goal is to replace the implementation, in other words call the library from `drawLineOnIjs` and `drawLineOnMask`

i created `drawLineOnIjs` using `bresenham-zingl`

package.json
```diff
@@ -58,6 +58,7 @@
     "@types/robust-point-in-polygon": "^1.0.2",
     "@vitejs/plugin-react": "^1.3.2",
     "autoprefixer": "^10.4.7",
+    "bresenham-zingl": "^0.1.1",
```
targos, line 61: this is not a development dependency.

src/draw/drawLineOnIjs.ts
```diff
@@ -42,37 +44,39 @@ export function drawLineOnIjs(
...
-      } else {
-        err += dColumn;
-        row += rowIncrement;
-      }
-    }
+  const numberChannels = Math.min(newImage.channels, color.length);
+  if (
+    [ImageColorModel.GREYA, ImageColorModel.RGBA].includes(image.colorModel)
+  ) {
+    lineAA(
```
targos, line 52: Anti-aliasing will be an option in the fututre but we don't want it for now

wadjih-bencheikh18: ok removed

package.json
wadjih-bencheikh18: added to dependencies

targos:
Lgtm. Can you check the CI errors?

targos:
OK I know what the problem is. Working on it.

PR #124, opened 2022-07-04
feat: implement drawLine on Mask with bresenham-zingl

wadjih-bencheikh18:
closes : https://github.com/image-js/image-js-typescript/issues/125?notification_referrer_id=NT_kwDOA2I9VLMzOTMyNDY0NjY4OjU2NzY5ODc2#issuecomment-1173807865


wadjih-bencheikh18:
the test `roi/getConvexHullMask/random shape, filled = false` is now failing

targos:
@wadjih-bencheikh18 Océane doesn't work this week. Can you look into why it fails? Maybe the new behavior is better than before.

wadjih-bencheikh18:
> @wadjih-bencheikh18 Océane doesn't work this week. Can you look into why it fails? Maybe the new behavior is better than before.

yes ofc no problem

src/draw/drawLineOnIjs.ts
```diff
@@ -40,13 +40,15 @@ export function drawLineOnIjs(
...
     bitDepth: [8, 16],
   });
+  const numberChannels = Math.min(newImage.channels, color.length);
 
   if (from.column === to.column && from.row === to.row) {
+    for (let channel = 0; channel < numberChannels; channel++) {
+      newImage.setValue(from.column, from.row, channel, color[channel]);
+    }
     return newImage;
   }
```
targos, line 50: I think you can just remove the `if` and let the library handle this case?

src/draw/drawLineOnMask.ts
```diff
@@ -29,6 +29,7 @@ export function drawLineOnMask(
   const newMask = maskToOutputMask(mask, options, { clone: true });
 
   if (from.column === to.column && from.row === to.row) {
+    newMask.setBit(from.column, from.row, 1);
     return newMask;
   }
```
targos, line 34: same

src/draw/drawLineOnIjs.ts
wadjih-bencheikh18: yes okey no problem

wadjih-bencheikh18:
> the test `roi/getConvexHullMask/random shape, filled = false` is now failing @opatiny

about this 
i found that there's an example that return diffrent line between the library algo and our algorithm

line (0,2) to (4,0)
```
[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0],
```


result of our algo
```
[0, 0, 0, 0, 1],
[0, 0, 1, 1, 0],
[1, 1, 0, 0, 0],
```

result of the library
```
[0, 0, 0, 1, 1],
[0, 1, 1, 0, 0],
[1, 0, 0, 0, 0],
```

src/draw/drawLineOnIjs.ts
wadjih-bencheikh18: yes it handle the case

targos:
The new result looks good to me (I think both are technically correct). @lpatiny ?

PR #126, opened 2022-07-04
feat: implement drawCircle using bresenham-zingl

wadjih-bencheikh18:
after merging https://github.com/image-js/image-js-typescript/pull/124 (same point line case) tests will run correctly

also i think filling the circle can be optimized, if we do it symmetrically

another point
the library is using 4 symetric points to draw the circle, and in the previous algo i used 8 points 
image

### DIFF
--- a/src/draw/drawCircleOnIjs.ts
+++ b/src/draw/drawCircleOnIjs.ts
@@ -1,3 +1,5 @@
+import { circle } from 'bresenham-zingl';
+
 import { IJS } from '../IJS';
 import checkProcessable from '../utils/checkProcessable';
 import { getDefaultColor } from '../utils/getDefaultColor';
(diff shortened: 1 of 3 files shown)


src/draw/drawCircleOnIjs.ts
```diff
@@ -43,71 +45,44 @@ export function drawCircleOnIjs(
...
-    newImage.setPixel(-column + cColumn, row + cRow, color);
-    newImage.setPixel(row + cColumn, -column + cRow, color);
-    newImage.setPixel(-row + cColumn, -column + cRow, color);
-    newImage.setPixel(-column + cColumn, -row + cRow, color);
-    // }
-    if (fill) {
-      fillCircle(column, row);
-    }
+  if (radius < 0) {
+    throw Error('circle radius must be positive');
```
targos, line 53: please add a test for this case

wadjih-bencheikh18: done thank you

PR #131, opened 2022-07-13
Implement `cropAlpha`

opatiny:


src/roi/__tests__/getMask.test.ts
```diff
@@ -145,4 +145,17 @@ describe('getMask', () => {
...
+      [0, 1, 1],
+      [0, 0, 1],
+      [0, 1, 1],
+    ]);
+    const roiMapManager = fromMask(mask);
+
+    const roi = roiMapManager.getRois({ kind: RoiKind.WHITE })[0];
+    const roiMask = roi.getMask();
+
+    expect(roiMask.origin).toStrictEqual({ column: 1, row: 0 });
```
targos, line 159:
```suggestion
    expect(roiMask.origin).toStrictEqual(roi.origin);
```

src/operations/cropAlpha.ts
```diff
@@ -0,0 +1,99 @@
+import { IJS } from '../IJS';
+import checkProcessable from '../utils/checkProcessable';
+
+export interface CropAlphaOptions {
+  /**
+   * Threshold above which rows and columns should be kept.
```
targos, line 6:
```suggestion
   * Threshold from which rows and columns should be kept.
```

The default is maxValue and there is nothing above it.

PR #135, opened 2022-07-14
Implement `median` and `average`

opatiny:
- refactor(floodFill): use origin
- refactor(copyTo): use origin
- test(cropAlpha): fix coverage
- docs(demo): fix copyTo demo
- fix(createFrom): copy input image origin
- fix: copy origin in convertDepth and extract
- feat: implement getChannel
- feat: implement average
- fix: fix getChannel
- feat: implement median


src/operations/extract.ts
```diff
@@ -43,6 +43,7 @@ export function extract(
 
   let newImage = new IJS(resultWidth, resultHeight, {
     colorModel: image.colorModel,
+    origin: image.origin,
```
targos, line 46: Let's use `origin` from the options instead.

Issue #142, opened 2022-07-26
Allow threshold as percent


targos:
It was possible in legacy: https://github.com/image-js/image-js/blob/b3b7499a5b0e6ee93f69e0646115653b3c044687/src/util/converter.js#L25

PR #150, opened 2022-09-22
feat(threshold): authorize threshold in percents

opatiny:
Closes: #142

### DIFF
--- a/src/operations/threshold.ts
+++ b/src/operations/threshold.ts
@@ -46,9 +46,9 @@ interface ThresholdOptionsBase {
 
 export interface ThresholdOptionsThreshold extends ThresholdOptionsBase {
   /**
-   * Threshold value that should be used. Should be an integer between 0 and Image.maxValue.
+   * Threshold value that should be used. Should be an integer between 0 and Image.maxValue or a value in percents as a string, like "40%".
    */
-  threshold: number;
+  threshold: number | string;
 }
 
 export interface ThresholdOptionsAlgorithm extends ThresholdOptionsBase {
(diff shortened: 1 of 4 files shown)


src/operations/threshold.ts
```diff
@@ -129,7 +129,24 @@ export function computeThreshold(
...
+    const threshold = options.threshold;
+    if (typeof threshold === 'number') {
+      thresholdValue = threshold;
+    } else if (
+      typeof threshold === 'string' &&
+      threshold.endsWith('%') &&
+      !isNaN(Number(threshold.slice(0, threshold.length - 1)))
+    ) {
+      const percents = Number(threshold.slice(0, threshold.length - 1));
+      if (!(percents >= 0 && percents <= 100)) {
```
targos, line 141:
```suggestion
      if (percents < 0 || percents > 100) {
```

src/operations/threshold.ts
```diff
@@ -129,7 +129,24 @@ export function computeThreshold(
...
+        throw new Error(
```
targos, line 142:
```suggestion
        throw new RangeError(
```

src/utils/validators.ts
```diff
@@ -33,9 +33,9 @@ export function validateChannel(channel: number, image: Image): void {
  * @param image - Image from which the value comes.
  */
 export function validateValue(value: number, image: Image): void {
```
targos, line 35: I think `validateValue` is used in other places where we really want the value to be an integer?

opatiny: It's apparently only used in `fill` and `fillAlpha` when the user passes data to be put directly in the image. Since `data` is a typed array, shouldn't it clamp the values automatically?

opatiny: I added a test and the values are indeed clamped. I don't know then if we want to force the user to input integers?

targos: ok

PR #153, opened 2022-10-18
Implement comparison functions

opatiny:
- feat: impelement meanDifference
- test: add meanDifference tests
- feat: implement mse and rmse
- test: add rmse testcases
- feat: implement psnr
- fix(mean): allow to compute mean on all values
- feat: implement variance
- feat: implement validateForComparison
- feat: implement ssim
- fix: fix compute functions parameters
- feat: implement computeDssim
- fix: computeSsim works
- docs(demo): add computeSsim demo
- fix(computeSsim): add algorithm option
- test: add computeDssim test


src/compare/computeMeanDifference.ts
```diff
@@ -0,0 +1,34 @@
+import { Image } from '..';
+
+/**
+ * Compute the mean difference between two images.
+ * Ignore values that are zero in both images.
```
targos, line 5: Do you have a reference for this behavior?

src/compare/__tests__/computeSsim.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  expect(computeMse(original, other)).toBe(0);
+});
+
+// In the following tests the SSIM expected values were computed with Matlab.
+
+test('more contrast', async () => {
+  let original = testUtils.load('ssim/ssim-original.png');
+  let other = testUtils.load('ssim/ssim-contrast.png');
+  expect(computeSsim(original, other).mssim).toBeCloseTo(0.8201, 1);
+});
```
targos, line 51: Did you find the problem you hade with the comparison to Matlab?

src/compare/computeMeanDifference.ts
opatiny: As this is not a commonly used algorithm, it is removed from `image-js`. The algorithm was copied to this issue: https://github.com/cheminfo/profid-padif/issues/20

src/compare/__tests__/computeSsim.test.ts
opatiny: Yes, I was passing GREY images, whereas the library only accepts RGBA images....

PR #162, opened 2022-11-08
Implement FAST keypoints

opatiny:
- fix: drawCircle options were not optional
- feat: implement getCirclePoints
- feat: implement isFastKeypoints
- fix(isFastKeypoint): all compare values the same
- feat: implement getCornerScore
- fix: isFastKeypoint uses predefined points
- refactor: rename getCornerScore to getKeypointScore
- fix(getFastKeypoints): remove duplicate keypoints
- test: add getFastKeypoints tests
- fix(getFastKeypoints): add nonMaxSuppresssion option


src/utils/getCirclePoints.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Point } from '../geometry';
+
+/**
+ * Get the coordinates of the points on a circle. The reference is the center of the circle.
+ * Caution: the points are not sorted.
+ *
+ * @param radius - Radius of the circle.
+ * @returns The coordinates of the points on a circle of given diameter.
+ */
+export function getCirclePoints(radius: number): Point[] {
```
targos, line 12: You created this utility but it isn't used anywhere in the code.

src/featureMatching/getFastKeypoints.ts
```diff
@@ -0,0 +1,115 @@
...
+        })
+      ) {
+        possibleCorners.push({ row, column });
+      }
+    }
+  }
+
+  const allKeypoints: FastKeypoint[] = [];
+
+  // Non-Maximal Suppression
```
targos, line 77: Is this comment in the wrong place? Asking because the check for `!nonMaxSuppression` is done later.

src/featureMatching/isFastKeypoint.ts
```diff
@@ -0,0 +1,127 @@
...
+      } else {
+        currentLength++;
+      }
+    } else {
+      counterArray.push(currentLength);
+      currentLength = 1;
+    }
+  }
+
+  if (Math.max(...counterArray) >= nbContiguousPixels) {
```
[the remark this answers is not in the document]
opatiny: It should remain a lot smaller than 100 values.

src/utils/getCirclePoints.ts
opatiny: I would actually need it, but I didn't know how to implement it properly. I need the points on a circle ordered, but my current implementation returns the points mixed up. Do you have an advice on how to get the coordinates of pixels on a circle with given radius?

src/utils/getCirclePoints.ts
```diff
@@ -18,3 +18,25 @@ export function getCirclePoints(radius: number): Point[] {
...
 }
+
+/**
+ * Get the coordinates of the points that are on right, bottom, left and top at a given radius. The reference is the center of the circle.
+ * First point is the most on the right, then points are in clockwise order.
+ *
+ * @param radius - Radius of the circle.
+ * @returns The coordinates of the compass points.
+ */
+export function getCompassPoints(radius: number): Point[] {
```
targos, line 29: This is a new unused (and untested) function.

src/utils/getCirclePoints.ts
```diff
@@ -0,0 +1,53 @@
...
+  circle(radius, radius, radius, (column: number, row: number) => {
+    circlePoints.push({ row: row - radius, column: column - radius });
+  });
+
+  let firstQuarter: Point[] = [];
+  let secondQuarter: Point[] = [];
+  let thirdQuarter: Point[] = [];
+  let fourthQuarter: Point[] = [];
+
+  for (let i = 0; i < circlePoints.length / 4; i++) {
```
targos, line 24: is `circlePoints.length` guaranteed to be a multiple of 4?

targos: You could also do `for (let i = 0; i < circlePoints.length; i+=4) {` to avoid having to multiply by 4 in the body of the loop.

src/featureMatching/getFastScore.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Point } from '../geometry';
+
+/**
+ * Compute the score of a keypoint using the function described in the FAST article.
+ * DOI: https://doi.org/10.1007/11744023_34
+ *
+ * @param origin - Keypoint coordinates.
+ * @param image - Image to process
+ * @param threshold - FAST threshold.
+
```
targos, line 11: Can you fix the ESLint warnings here?

src/utils/getCirclePoints.ts
opatiny: > is `circlePoints.length` guaranteed to be a multiple of 4?

I think it should be, the algorithm generating the circle points creates the points of a quarter of circle and then uses symmetries to get the three remaining quarters.

Issue #176, opened 2022-12-06
Generate feature descriptors with BRIEF


opatiny:
We want the descriptors to take orientation into account -> rBRIEF.

We want to generate a unique string characterising each keypoint. For BRIEF, it consists of a 256 bits string. The bits are the result of the intensity comparison of two pixels within the window around the keypoint. To make the features rotation invariant, the window should first be rotated, and then the comparisons can be made.

The ORB article, in which the computation of the descriptors can be found: http://www.gwylab.com/download/ORB_2012.pdf
DOI: [10.1109/ICCV.2011.6126544](https://doi.org/10.1109/ICCV.2011.6126544)

A video basically explaining the same thing in a more understandable way (from 37:05 to the end): https://www.youtube.com/watch?v=CMolhcwtGAU

opatiny:
Now I have the following issue, which I don't know how to solve: we want to find the best 256 comparisons between two pixels of the patch around the keypoint that will result in descriptors as different as possible.

Apparently, this is commonly done using machine learning, and is depending on the type of images used, so a training set is necessary.

opatiny:
Based on the BRIEF article, we can see that the points chosen are not that important as long as we use systematically the same ones. We decide to use points with a gaussian distribution around the center of the image.

**BRIEF article:** https://www.cs.ubc.ca/~lowe/525/papers/calonder_eccv10.pdf
DOI: https://doi.org/10.1007/978-3-642-15561-1_56

PR #181, opened 2022-12-13
Implement rotated BRIEF descriptors

opatiny:
- feat: implement compareIntensity
- wip: start working on getBriefDescriptors
- docs: enhance rotate doc
- docs: enhance moments and centroid doc
- fix(getBriefDescriptors): extract keypoint patch
- feat: add generic clamp function
- feat: implement getGaussianPoints
- fix: change getGaussianPoints parameters
- feat: implement extractSquareImage
- feat: implement getBriedDescriptors
- fix(getBriefDescriptors): tests run until the end
- fix(getBriefDescriptors): avoid even dimensions
- test: visualise default points pairs


targos:
Does it close any issues?

src/featureMatching/__tests__/__snapshots__/getBriefDescriptors.test.ts.snap
```diff
@@ -0,0 +1,2849 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`alphabet image, maxNbFeatures = 10 2`] = `
```
targos, line 3: Snapshot tests aren't very useful if we can't review them. Unless you don't care about the actual values and just want to prevent future regressions?

src/utils/clamp.ts
```diff
@@ -37,3 +37,6 @@ function clamp255(value: number): number {
 function clamp65535(value: number): number {
   return Math.min(Math.max(value, 0), 65535);
 }
+
+export const clamp = (min: number, max: number) => (value: number) =>
+  Math.max(Math.min(value, max), min);
```
targos, line 42: Please use regular named functions here. Also please rename the exported one so it's clear that it returns a function and not a clamped value.
Btw, any reason for using the opposite order of Math functions compared to the above implementations?

opatiny: What do you mean by "regular named functions"?

I used this syntax, which is apparently called "currying" because I saw it online and found it interesting. Do you think it's relevant to keep it or should I use a more standard syntax?

No, there is no reason for the order of the math functions, I just didn't pay attention.

src/featureMatching/__tests__/__snapshots__/getBriefDescriptors.test.ts.snap
opatiny: I have indeed used them to prevent regressions, because it is code that is difficult to test otherwise.

src/utils/clamp.ts
targos: I mean something like this:

```ts
export function getClampFromTo(min: number, max: number) {
  return function clamp(value: number) {
    return Math.max(Math.min(value, max), min);
  };
}
```

This is still called currying (which is not a syntax). The syntax that you used is "arrow functions", which are fine as small callbacks for example in `array.map` but should be avoided in exported places or especially when returned from other functions because they are anonymous and this is always annoying in stack traces or when debugging.

src/featureMatching/getBriefDescriptors.ts
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Size of the patch around the keypoint used to compute the descriptor.
+   *
+   * @default 31
+   */
+  patchSize?: number;
+  /**
+   * Number of bits of the final descriptor.
+   *
+   * @default 512
```
targos, line 32: the default is 256 in the code.

src/utils/clamp.ts
opatiny: > Does it close any issues?

Yes, #176

src/featureMatching/utils/getGaussianPoints.ts
```diff
@@ -84,7 +84,7 @@ export function getGaussianValues(
 
   const result = array.map((value) => {
     const rounded = Math.round(value);
-    return clamp(-width, width)(rounded);
+    return getClampFromTo(-width, width)(rounded);
```
targos, line 87: You can avoid recreating the function for every value if you put it in a variable before calling array.map.

src/utils/clamp.ts
```diff
@@ -38,5 +38,15 @@ function clamp65535(value: number): number {
...
 }
 
-export const clamp = (min: number, max: number) => (value: number) =>
-  Math.max(Math.min(value, max), min);
+/**
+ * Clamp a value to a given range.
+ *
+ * @param min - Lower threshold.
+ * @param max - Upper threshold.
+ * @returns The clamped value.
```
targos, line 46: This returns a function, not the clamped value.

PR #187, opened 2022-12-29
Implement brute force matching

opatiny:
- feat: implement hamming distance
- feat: implement bruteForceMatch
- feat: implement drawMatches
- fix: update featureMatching index
- fix: add bruteForceMatch options
- fix(getBriefDescriptors): handle edge cases
- feat: implement checkBorderDistance
- test(drawMatches): slightly rotated image runs
- fix: modify patch size in getBriefDescriptors
- chore: remove outdated image snapshots
- test: enhance getFastKeypoints test
- fix(drawMatches): add option to show keypoints


src/featureMatching/__tests__/getOrientedFastKeypoints.test.ts
```diff
@@ -1,6 +1,6 @@
 import { getOrientedFastKeypoints } from '../getOrientedFastKeypoints';
 
-test('7x7 image, angle = 90°', () => {
+test.only('7x7 image, angle = 90°', () => {
```
targos, line 3: !

opatiny: Could we add an `eslint` rule or another check that would verify that there is no `test.only` in a file and either throw an error or a warning if there is? I find it easy to forget one while debugging.

src/operations/copyTo.ts
```diff
@@ -42,7 +42,9 @@ export function copyTo(
   const { column, row } = origin;
 
   if (source.colorModel !== target.colorModel) {
-    throw new Error('Source and target should have the same color model.');
+    throw new Error(
+      'copyTo: source and target must have the same color model.',
+    );
```
targos, line 47: Any reason for including the name of the function in the error message? this is not a common pattern in JS, since the name is always visible in the first line of the stack trace.

src/featureMatching/__tests__/getOrientedFastKeypoints.test.ts
targos: There is an eslint plugin for Jest: https://github.com/jest-community/eslint-plugin-jest#readme
We used to depend on it by default but it's not the case anymore. You can try to install it in this package and extend the recommended config

targos: The relevant rule is https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-disabled-tests.md

src/operations/copyTo.ts
opatiny: I don't know why, but I have been using this syntax a lot in the whole project, would you like me to remove all the function names? I think some of the already existing messages were like that and I replicated it.

Do we have a standard for whether the error messages should start with a lower or uppercase?

What should I do with the validators error messages, which contains the `process` parameter?

targos: > would you like me to remove all the function names? 

I think we should make an issue for this, yes

> Do we have a standard for whether the error messages should start with a lower or uppercase?

It should preferably start with an uppercase

> What should I do with the validators error messages, which contains the process parameter?

Remove the parameter.

opatiny: Ok, I'll create an issue then

opatiny: https://github.com/image-js/image-js-typescript/issues/192

targos (approved):
feel free to merge when you want

src/featureMatching/__tests__/getOrientedFastKeypoints.test.ts
opatiny: > The relevant rule is https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-disabled-tests.md

It seems like that rule doesn't cover the case we're interested in:
image

targos: you're right. I forgot the name of the rule you want, but it exists.

opatiny: Ok from what I see this would be the right rule: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-focused-tests.md

opatiny: Could you help me out with the syntax in `.eslintrc.yml`?

I did something like this, but it doesn't work.

```yml
root: true
extends: [cheminfo-typescript, cheminfo-typescript/jsdoc]
plugins: [jest]
rules:
  '@typescript-eslint/restrict-template-expressions': off
  'jest/no-focused-tests': 'warn'
```

opatiny: I wasn't sure about the syntax for `.yml` files and eslint wouldn't run, but apparently I had just forgotten to install a package. It works now, sorry

opatiny: I had to install `yarn` to install the package, and it crated a `yarn.lock` file. Is this file useful?

targos: > I had to install yarn to install the package

Yarn is an alternative package manager to npm. There shouldn't be a need to install it.

PR #226, opened 2023-02-28
chore: implement browser versions for node-only api methods

stropitek:


stropitek:
Idea: implement the read function for the browser:

```ts
async function read(dataOrUrl: string | URL | File): Image {
  // If File use FileReader + decode
  // If string or URL use fetch + decode
}
```

Not sure wether it should be 2 separate functions because it could be convenient to have a sync function when having a `File` as input.

targos:
Since `fetch` is also supported in Node.js, I think we should have a different method for it.

targos:
I don't think we can read a `File` synchronously?

targos (approved):

stropitek:
> I don't think we can read a `File` synchronously?

You are right, I thought that `onload` on `FileReader` was called synchronously

PR #237, opened 2023-03-10
feat!: add features that calculate convexHull, borders length etc.

EscapedGibbon:


src/roi/__tests__/ratio.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    [0, 1, 1, 1],
+    [0, 1, 0, 1],
+    [0, 1, 1, 1],
+    [0, 0, 0, 0],
+  ]);
+
+  expect(mask.ratio).toBe(4 / 4);
+});
+
+test('ratio of ROI', () => {
```
targos, line 12: Different tests should have different names.
The test name should be more precise about what it tests.

src/roi/__tests__/ratio.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  const mask = testUtils.createMask([
+    [0, 1, 1, 1, 0, 0, 0],
+    [0, 1, 0, 1, 0, 0, 0],
+    [0, 1, 1, 1, 0, 0, 0],
+    [0, 0, 0, 0, 0, 0, 0],
+    [0, 0, 0, 0, 0, 0, 0],
+  ]);
+  expect(mask.ratio).toBe(7 / 5);
```
targos, line 20: @stropitek didn't you tell me that the goal was to implement the properties on ROI objects? This test seems to imply that it's on a Mask.

EscapedGibbon: My bad, I wanted to test the ratio and unintentionally put the mask ratio test into ROI testing folder.

src/Mask.ts
```diff
@@ -512,12 +512,16 @@ export class Mask {
...
-   * @returns Array of boder points.
+   * @returns Array of border points.
    */
   public getBorderPoints(options?: GetBorderPointsOptions): Point[] {
     return getBorderPoints(this, options);
   }
 
+  get ratio() {
+    return this.width / this.height;
+  }
```
stropitek, line 523: I think you can remove this for now.

If it's implemented is should be called `aspectRatio` and done for all types of images, not just masks.

You can keep it on ROIs.

stropitek (commented):
I reviewed the code and here are a few notes I took.

@EscapedGibbon please don't do anything yet regarding those, we need to discuss this first with @lpatiny and @targos 

- There are a lot of places where the same iteration logic is repeated: iterate over ROI box, calculate data index, access data. Maybe implement a generator.
- The same index calculation is done in many property computers. It is a perf issue?
- Lot of repeated logic for cached computed properties
- Public properties: minX, surface etc... They are intrinsic properties of the ROI that never change, and they should be read-only. But for that they need to be passed right away in the constructor, not modified later.
- map, getMap should be hidden behind a Symbol? Seems like an internal thing that should not be exposed to the lib user.
- Definitions: what is an internal ID, an external ID, box IDs?
- References: is the algo to get borders published / documented somewhere?

lpatiny:
In this new project we replaced 'x' by 'column' and 'y' by 'row'. Please rename the variables to be consistent. Also use `maxColumn`, `minRow`, etc.

@stropitek Link to border ? You mean how we calculate the perimeter ?

We have the documentation in our help about which method we are using: https://docs.c6h6.org/docs/eln/uuid/187642421e225f8cfcd65a9243394a72

That also contains the link to the review of the literature exploring various methods: https://www.duo.uio.no/bitstream/handle/10852/9541/LYang-3.pdf?sequence=1&isAllowed=y

src/roi/Roi.ts
```diff
@@ -31,7 +37,34 @@ export class Roi {
...
+  public maxX: number;
+  public maxY: number;
+  public minX: number;
+  public minY: number;
+
+  private computed: {
+    perimeter?: number;
+    borderIDs?: number[];
+    perimeterInfo?: { one: number; two: number; three: number; four: number };
+    BoxIDs?: number[];
```
lpatiny, line 51: I think this property does not exists (with the uppercase letter)

src/roi/Roi.ts
```diff
@@ -40,8 +73,12 @@ export class Roi {
     this.width = 0;
     this.height = 0;
     this.surface = 0;
+    this.computed = {};
+    this.minX = Infinity;
+    this.minY = Infinity;
+    this.maxX = -Infinity;
+    this.maxY = -Infinity;
```
lpatiny, line 80: I think those 4 parameters should disappear and are replaced by `origin` and `width`/ `height`

src/roi/computeRois.ts
```diff
@@ -64,24 +76,65 @@ export function computeRois(roiMapManager: RoiMapManager): void {
...
+          }
+          if (y < roi.minY) {
+            roi.minY = y;
+          }
+          if (y > roi.maxY) {
+            roi.maxY = y;
+          }
+        }
+      }
+    }
```
lpatiny, line 106: I'm convince maxX, maxY, minX, minY are useless and already in origin, width, height.

src/roi/computeRois.ts
```diff
@@ -64,24 +76,65 @@ export function computeRois(roiMapManager: RoiMapManager): void {
...
```
lpatiny, line 137: Same comment here.

src/roi/computeRois.ts
```diff
@@ -11,25 +11,37 @@ export function computeRois(roiMapManager: RoiMapManager): void {
 
   const whites = new Array(map.nbPositive);
   const blacks = new Array(map.nbNegative);
-
   for (let i = 0; i < map.nbPositive; i++) {
     whites[i] = {
+      borderLengths: [],
+      borderIDs: [],
```
lpatiny, line 17: Not sure we need to define those properties here.

lpatiny (changes requested):

src/Mask.ts
stropitek: Can you point me to the definition of "aspect ratio" being a value between 0 and 1?

When searching I'm consistently finding that it's "the relationship between width and height"

EscapedGibbon: @stropitek I think the aspectRatio of the mask is calculated as part of  feretDiameter. It is minDiameter/maxDiameter. It is calculated in getFeret() function which is in a maskAnalysis folder

targos: This discussion is about the aspect ratio of Mask/Image, not regions of interest.

src/roi/Roi.ts
```diff
@@ -91,4 +138,516 @@ export class Roi {
   public getBorderPoints(options?: GetBorderPointsOptions): Array<Point> {
     return getBorderPoints(this, options);
   }
+
+  _computeBorderIDs(): void {
+    let borders = getBorders(this);
+    this.computed.borderIDs = [...borders.ids];
```
lpatiny, line 144: I guess we could keep borders.ids and we don't need to create a new array here

In order to prevent to use this method we can do hte same as with `getBorders` and call `computeBorderIDs(this)`.

src/Mask.ts
lpatiny: Ah yes indeed. Didn't see there were changes in Mask as well. I though it was only on ROIs. I agree we can remove it for now.

src/roi/Roi.ts
```diff
@@ -91,4 +138,516 @@ export class Roi {
...
+}
+
+/**
+ *
+ * @param roi - ROI
+ * @returns the surface of holes in ROI
+ */
+function getHolesInfo(roi: Roi) {
+  let surface = 0;
+  let width = roi.getMap().width;
```
lpatiny, line 456: If the value will not change we should take the habbit to define it as `const`. This applies everywhere.

src/roi/Roi.ts
```diff
@@ -91,4 +138,516 @@ export class Roi {
...
+  let data = roi.getMap().data;
+  for (let x = 1; x < roi.width - 1; x++) {
+    for (let y = 1; y < roi.height - 1; y++) {
+      let target = (y + roi.minY) * width + x + roi.minX;
+      if (roi.internalIDs.includes(data[target]) && data[target] !== roi.id) {
+        surface++;
+      }
+    }
+  }
+  return {
```
lpatiny, line 466: Could you add a testcase on holesInfo.

src/roi/Roi.ts
```diff
@@ -91,4 +145,513 @@ export class Roi {
...
+
+  get roundness() {
+    /*Slide 24 https://static.horiba.com/fileadmin/Horiba/Products/Scientific/Particle_Characterization/Webinars/Slides/TE011.pdf */
+    return (
+      (4 * this.surface) /
+      (Math.PI * this.feretDiameters.maxDiameter.length ** 2)
+    );
+  }
+
+  get feretDiameters(): {
```
lpatiny, line 326: It does not only contain Diameters but also the aspectRatio. I would rename it to `feret`. Also the type should not be redefined. It exists and is returned by `getFeret` (type = `Feret`)

src/roi/Roi.ts
```diff
@@ -91,4 +145,513 @@ export class Roi {
...
+   */
+  toJSON() {
+    return {
+      id: this.id,
+      height: this.height,
+      width: this.width,
+      surface: this.surface,
+      eqpc: this.eqpc,
+      ped: this.ped,
+      feretDiameterMin: this.feretDiameters.minDiameter,
```
lpatiny, line 348: I would keep the hierarchy and export `feret: this.feret`

src/roi/Roi.ts
```diff
@@ -31,17 +37,47 @@ export class Roi {
   /**
    * Surface of the ROI.
    */
+
   public surface: number;
 
+  public maxColumn: number;
+  public maxRow: number;
```
lpatiny, line 44: I don't think those 2 properties should be defined. Currently they are used once in the full code and they can easily be calculated on the fly.

Even if there is a calculation in the end condition `for (let i=0; i<roi.origin.column+roi.width-1; i++)` V8 will likely optimize this and will do only once the calculation (https://jsbench.me/ailfjqjx7r/1)

src/roi/Roi.ts
```diff
@@ -91,4 +145,513 @@ export class Roi {
...
+  let roiMap = roi.getMap();
+  let data = roiMap.data;
+  let one = 0;
+  let two = 0;
+  let three = 0;
+  let four = 0;
+
+  for (let x = 0; x < roiMap.width; x++) {
+    for (let y = 0; y < roiMap.height; y++) {
+      let target = (y + roi.origin.row) * roiMap.width + x + roi.origin.column;
```
lpatiny, line 382: I'm surprise that is is done on the full roiMap and not just target the specific ROI. This must be a bug and the current limiting step of image analysis.

lpatiny: I'm surprise that is is done on the full roiMap and not just target the specific ROI. This must be a bug and the current limiting step of image analysis.

src/roi/Roi.ts
```diff
@@ -22,55 +42,42 @@ export class Roi {
...
 
-  public constructor(map: RoiMap, id: number) {
+  #computed: Partial<Computed>;
+
+  public constructor(
+    map: RoiMap,
+    id: number,
+    width: number,
+    height: number,
+    origin: { row: number; column: number },
```
targos, line 72: I think we have a `Point` type for this

src/roi/Roi.ts
```diff
@@ -91,4 +138,584 @@ export class Roi {
...
+  let roiMap = roi.getMap();
+  let data = roiMap.data;
+  let one = 0;
+  let two = 0;
+  let three = 0;
+  let four = 0;
+
+  for (let column = 0; column < roi.origin.column + roi.width; column++) {
+    for (let row = 0; row < roi.origin.row + roi.height; row++) {
+      let target = computeIndex(roi, row, column);
```
lpatiny, line 439: I don't think those loop end values are correct if you are using computeIndex.

src/roi/Roi.ts
```diff
@@ -91,4 +138,584 @@ export class Roi {
...
+    );
+  }
+
+  get points() {
+    return this.#getComputed('points', () => {
+      if (!this.#computed.points) {
+        let points = [];
+        for (let row = 0; row < this.height; row++) {
+          for (let column = 0; column < this.width; column++) {
+            let target =
```
lpatiny, line 219: Why not using the helper here ?

src/roi/Roi.ts
```diff
@@ -91,4 +138,584 @@ export class Roi {
...
+  let ids: number[] = Array.from(surroudingIDs);
+  let borderLengths = ids.map((id) => {
+    return surroundingBorders.get(id);
+  });
+  return {
+    ids,
+    lengths: borderLengths,
+  };
+}
+function computeIndex(roi: Roi, y: number, x: number): number {
```
lpatiny, line 718: I would add this method in the class so that we have access to 'this' and can simplify the code. This method can be public I think.
@stropitek WDYT ?

src/roi/Roi.ts
lpatiny: I fixed this issue and allowed. In my screw project it allows to calculate all the paramters in 2.12s instead of 5.47s. I'm checking now what is the next limiting speed step.

src/roi/Roi.ts
```diff
@@ -91,4 +138,584 @@ export class Roi {
...
+      info.two * 2 +
+      info.three * 3 +
+      info.four * 4 -
+      delta * (info.two + info.three * 2 + info.four)
+    );
+  }
+
+  get points() {
+    return this.#getComputed('points', () => {
+      if (!this.#computed.points) {
```
stropitek, line 215: The functions that get passed to `#getComputed` should unconditionally compute the property and return it. It should not care about the `#computed` property.

src/roi/Roi.ts
```diff
@@ -91,4 +138,584 @@ export class Roi {
...
+  }
+
+  get holesInfo() {
+    return this.#getComputed('holesInfo', () => {
+      return getHolesInfo(this);
+    });
+    // if (!this.#computed.holesInfo) {
+    //   this.#computed.holesInfo = getHolesInfo(this);
+    // }
+    // return this.#computed.holesInfo;
```
stropitek, line 269: Please remove all commented code

src/roi/Roi.ts
```diff
@@ -91,4 +138,452 @@ export class Roi {
...
+            column +
+            this.origin.column;
+          if (this.map.data[target] === this.id) {
+            points.push([column, row]);
+          }
+        }
+      }
+      this.#computed.points = points;
+
+      return this.#computed.points;
```
stropitek, line 209:
```suggestion
      return points;
```

src/roi/Roi.ts
```diff
@@ -91,4 +138,452 @@ export class Roi {
   public getBorderPoints(options?: GetBorderPointsOptions): Array<Point> {
     return getBorderPoints(this, options);
   }
+
+  _computeBorderIDs(): { ids: number[]; lengths: number[] } {
+    const borders = getBorders(this);
+    this.#computed.borderIDs = borders.ids;
+    this.#computed.borderLengths = borders.lengths;
+    return borders;
```
stropitek, line 146: The ids and length should be in one computed property which returns an array of `{id: number, length: number}`

Todo in another PR

src/roi/Roi.ts
```diff
@@ -91,4 +138,452 @@ export class Roi {
...
+   */
+  get solidity() {
+    return this.surface / getConvexHull(this.getMask()).surface;
+  }
+
+  get convexHull() {
+    return this.#getComputed('convexHull', () => {
+      return getConvexHull(this.getMask());
+    });
+  }
```
stropitek, line 292: Should be refactored to not need creating a new Mask

src/roi/Roi.ts
```diff
@@ -91,4 +138,452 @@ export class Roi {
...
+
+  get mbr() {
+    return this.#getComputed('mbr', () => {
+      return getMbr(this.getMask());
+    });
+  }
```
stropitek, line 298: Should be refactored to not need creating a new Mask

src/roi/Roi.ts
```diff
@@ -91,4 +138,452 @@ export class Roi {
...
+      this.#computed[property] = result;
+      return result;
+    }
+    return this.#computed[property] as Computed[T];
+  }
+
+  computeIndex(y: number, x: number): number {
+    const roiMap = this.getMap();
+    return (y + this.origin.row) * roiMap.width + x + this.origin.column;
+  }
```
stropitek, line 349: Make this private

TODO in another PR

stropitek (commented):
@EscapedGibbon you can work on the comments that don't say that it's for another PR before we merge it.

Can you confirm that you incorporated the tests from image-js legacy as they were and did not change their behaviour?

EscapedGibbon:
@stropitek By incorporating the tests, do you mean that tested figures must give the same values as in legacy or that the test suites must be exactly the same as in the legacy?

lpatiny:
> @stropitek By incorporating the tests, do you mean that tested figures must give the same values as in legacy or that the test suites must be exactly the same as in the legacy?

The values should be the same but obviously some property names changed. If there were some changes comparing to the legacy code please provide and example.

EscapedGibbon:
> @EscapedGibbon you can work on the comments that don't say that it's for another PR before we merge it.
> 
> Can you confirm that you incorporated the tests from image-js legacy as they were and did not change their behaviour?

@stropitek Yes, I can confirm it

stropitek (approved):

Issue #262, opened 2023-03-30
ROI: Change shape of external and internal IDs


stropitek:
They should both be an array of `{id: number, length: number}`

_Originally posted by @stropitek in https://github.com/image-js/image-js-typescript/pull/237#discussion_r1153120105_

PR #276, opened 2023-04-06
chore: change externalIDs and borderIDs into arrays of objects with id and length properties

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/262

### DIFF
--- a/src/roi/Roi.ts
+++ b/src/roi/Roi.ts
@@ -13,13 +13,13 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 
-interface ID {
-  id: number;
+interface Border {
+  connectedID: number; // refers to the roiID of the contiguous ROI
   length: number;
 }
 interface Computed {
   perimeter: number;
-  borderIDs: ID[];
+  borders: Border[]; // external and internal ids which are not equal to roi ID
   perimeterInfo: { one: number; two: number; three: number; four: number };
   externalLengths: number[];
   borderLengths: number[];
@@ -30,7 +30,7 @@ interface Computed {
   boxIDs: number[];
   eqpc: number;
   ped: number;
-  externalIDs: ID[];
+  externalBorders: Border[];
   roundness: number;
   convexHull: { points: Point[]; surface: number; perimeter: number };
   mbr: Mbr;
@@ -158,8 +158,8 @@ export class Roi {
   /**
    * Return an array of ROIs IDs that touch the current ROI.
    */
-  get externalIDs(): ID[] {
-    return this.#getComputed('externalIDs', () => {
+  get externalBorders(): Border[] {
+    return this.#getComputed('externalBorders', () => {
       return this.getExternalIDs();
     });
   }
@@ -225,27 +225,30 @@ export class Roi {
     });
   }
 
-  getExternalIDs(): ID[] {
+  getExternalIDs(): Border[] {
     // take all the borders and remove the internal one ...
-    let borders = this.borderIDs;
+    let borders = this.borders;
 
     this.#computed.externalLengths = [];
-    this.#computed.externalIDs = [];
+    this.#computed.externalBorders = [];
 
     let internals = this.internalIDs;
 
     for (let border of borders) {
-      if (!internals.includes(border.id)) {
-        const element: ID = { id: border.id, length: border.length };
+      if (!internals.includes(border.connectedID)) {
+        const element: Border = {
+          connectedID: border.connectedID,
+          length: border.length,
+        };
 
-        this.#computed.externalIDs.push(element);
+        this.#computed.externalBorders.push(element);
       }
     }
-    const externalIDs = this.#computed.externalIDs;
+    const externalIDs = this.#computed.externalBorders;
     return externalIDs;
   }
-  get borderIDs() {
-    return this.#getComputed('borderIDs', () => {
+  get borders() {
+    return this.#getComputed('borders', () => {
       return getBorders(this);
     });
   }
@@ -377,20 +380,26 @@ function getPerimeterInfo(roi: Roi) {
         let nbAround = 0;
         if (column === 0) {
           nbAround++;
-        } else if (roi.externalIDs.find((el) => el.id === data[target - 1])) {
+        } else if (
+          roi.externalBorders.find((el) => el.connectedID === data[target - 1])
+        ) {
           nbAround++;
         }
 
         if (column === roiMap.width - 1) {
           nbAround++;
-        } else if (roi.externalIDs.find((el) => el.id === data[target + 1])) {
+        } else if (
+          roi.externalBorders.find((el) => el.connectedID === data[target + 1])
+        ) {
           nbAround++;
         }
 
         if (row === 0) {
           nbAround++;
         } else if (
-          roi.externalIDs.find((el) => el.id === data[target - roiMap.width])
+          roi.externalBorders.find(
+            (el) => el.connectedID === data[target - roiMap.width],
+          )
         ) {
           nbAround++;
         }
@@ -398,7 +407,9 @@ function getPerimeterInfo(roi: Roi) {
         if (row === roiMap.height - 1) {
           nbAround++;
         } else if (
-          roi.externalIDs.find((el) => el.id === data[target + roiMap.width])
+          roi.externalBorders.find(
+            (el) => el.connectedID === data[target + roiMap.width],
+          )
         ) {
           nbAround++;
         }
@@ -547,12 +558,12 @@ function getBoxIDs(roi: Roi): number[] {
  * @param roi - ROI
  * @returns borders' length and their IDs
  */
-function getBorders(roi: Roi): ID[] {
+function getBorders(roi: Roi): Border[] {
   const roiMap = roi.getMap();
   const data = roiMap.data;
   let surroundingID = 0;
   let surroundingBorder;
-  let surroundingIDs: ID[] = []; // allows to get a unique list without indexOf
+  let surroundingIDs: Border[] = []; // allows to get a unique list without indexOf
 
   let visitedData = new Set();
   let dx = [+1, 0, -1, 0];
@@ -588,16 +599,16 @@ function getBorders(roi: Roi): ID[] {
               }
 
               let index = surroundingIDs.findIndex(
-                (element) => element.id === data[neighbour],
+                (element) => element.connectedID === data[neighbour],
               );
               if (index === -1) {
                 surroundingIDs.push({
-                  id: surroundingID,
+                  connectedID: surroundingID,
                   length: surroundingBorder,
                 });
               } else {
                 surroundingIDs[index] = {
-                  id: surroundingID,
+                  connectedID: surroundingID,
                   length: surroundingBorder,
                 };
               }
--- a/src/roi/Roi.ts
+++ b/src/roi/Roi.ts
@@ -390,9 +390,7 @@ function getPerimeterInfo(roi: Roi) {
   let two = 0;
   let three = 0;
   let four = 0;
-  let externalIDs = roi
-    .getExternalBorders()
-    .map((element) => element.connectedID);
+  let externalIDs = roi.externalBorders.map((element) => element.connectedID);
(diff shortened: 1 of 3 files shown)


src/roi/Roi.ts
```diff
@@ -139,10 +143,9 @@ export class Roi {
     return getBorderPoints(this, options);
   }
   //TODO The ids and length should be in one computed property which returns an array of {id: number, length: number}
```
stropitek, line 145: Please remove the TODO

src/roi/Roi.ts
```diff
@@ -227,41 +230,37 @@ export class Roi {
...
-  get borderLengths() {
-    return this.#getComputed('borderLengths', () => {
-      return this._computeBorderIDs().lengths;
-    });
-  }
+  // get borderLengths() {
+  //   return this.#getComputed('borderLengths', () => {
+  //     return this._computeBorderIDs().lengths;
+  //   });
+  // }
```
stropitek, line 263: Remove commented code

src/roi/Roi.ts
```diff
@@ -139,10 +143,9 @@ export class Roi {
     return getBorderPoints(this, options);
   }
   //TODO The ids and length should be in one computed property which returns an array of {id: number, length: number}
-  _computeBorderIDs(): { ids: number[]; lengths: number[] } {
+  _computeBorderIDs(): ID[] {
```
stropitek, line 146: This methods should no longer exist

The logic should be in `get borderIDs`

src/roi/Roi.ts
```diff
@@ -227,41 +230,37 @@ export class Roi {
...
+
+        this.#computed.externalIDs.push(element);
       }
     }
     const externalIDs = this.#computed.externalIDs;
-    const externalLengths = this.#computed.externalLengths;
-    return { externalIDs, externalLengths };
+    return externalIDs;
   }
   //TODO Should be merged together in one borders property
```
stropitek, line 252: remove the TODO

src/roi/__tests__/borderLengths.test.ts
```diff
@@ -20,10 +20,12 @@ test('border lengths property 4x4', () => {
     [0, 1, 1, 1],
     [0, 0, 0, 0],
   ]);
+
   const roiMapManager = fromMask(mask);
+
   const rois = roiMapManager.getRois();
 
-  expect(rois[0].borderLengths).toStrictEqual([6, 1]);
+  expect(rois[0].borderIDs[0].length).toStrictEqual(6);
```
stropitek, line 28: You are no longer testing the same thing

src/roi/__tests__/borderLengths.test.ts
```diff
@@ -35,6 +37,6 @@ test('border lengths property 4x4', () => {
   ]);
   const roiMapManager = fromMask(mask);
   const rois = roiMapManager.getRois();
-  expect(rois[0].borderLengths).toStrictEqual([2, 1]);
-  expect(rois[1].borderLengths).toStrictEqual([2, 2]);
+  expect(rois[0].borderIDs[0].id).toStrictEqual(-2);
+  expect(rois[0].borderIDs[0].length).toStrictEqual(1);
```
stropitek, line 41: You are no longer testing the same thing

src/roi/Roi.ts
```diff
@@ -562,25 +567,33 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
-              if (!surroundingBorder) {
-                surroundingBorders.set(data[neighbour], 1);
+              surroundingID = data[neighbour];
+
+              if (surroundingBorder === undefined) {
+                surroundingBorder = 0;
               } else {
-                surroundingBorders.set(data[neighbour], ++surroundingBorder);
+                ++surroundingBorder;
+              }
```
stropitek, line 576: You significantly changed how this value is computed (it probably breaks `getBorders`)

Could you explain?

src/roi/__tests__/borderLengths.test.ts
EscapedGibbon: It still tests functionality of the borderLengths though. The way borderIDs look now is completely different so I had to change the tests.

src/roi/Roi.ts
EscapedGibbon: Before the surroundingBorder and surroundingIDs were easily taken from a map and a set respectively.  Now I can't do that because the set will not work with objects the way I want it to. Or maybe I just don't know how to implement it correctly. So I just calculate the values for the object separately and then create and object instead of picking specific values out of a set or a map.

src/roi/Roi.ts
```diff
@@ -13,9 +13,13 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 
+interface ID {
+  id: number;
```
lpatiny, line 17: connectedID // refers to the roiID of the contiguous ROI

src/roi/Roi.ts
```diff
@@ -26,7 +30,7 @@ interface Computed {
   boxIDs: number[];
   eqpc: number;
   ped: number;
-  externalIDs: number[];
+  externalIDs: ID[];
```
lpatiny, line 33: rename to `externalBorders`

src/roi/Roi.ts
```diff
@@ -13,9 +13,13 @@ import { RoiMap } from './RoiMapManager';
...
 import { getMask, GetMaskOptions } from './getMask';
 
+interface ID {
+  id: number;
+  length: number;
+}
 interface Computed {
   perimeter: number;
-  borderIDs: number[];
+  borderIDs: ID[];
```
lpatiny, line 22: rename to `borders`

Please confirm that this is internal + external and add a comment in 'getBorders'

src/roi/Roi.ts
```diff
@@ -26,7 +30,7 @@ interface Computed {
   boxIDs: number[];
```
lpatiny, line 30: Please check what is external and possibly remove it.

src/roi/Roi.ts
```diff
@@ -575,12 +576,15 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
     return surroundingBorders.get(id);
   });
-  return {
-    ids,
-    lengths: borderLengths,
-  };
+  let borders = id.map((id, index) => {
+    return {
+      connectedID: id,
+      length: borderLengths[index],
```
stropitek, line 586: you don't need `borderLengths` variable here.

you can `surroundingBorders.get(id)` here

src/roi/Roi.ts
```diff
@@ -189,7 +186,10 @@ export class Roi {
       delta * (info.two + info.three * 2 + info.four)
     );
   }
-
+  /**
+   * Points of the ROI
+   * Returns an array with arrays of 2 numbers which are the coordinates of all the current ROI points
```
stropitek, line 191:
```suggestion
   * An array of tuples, each tuple being the x and y coordinates of the ROI point.
```

src/roi/Roi.ts
```diff
@@ -221,48 +221,48 @@ export class Roi {
...
     return this.#getComputed('holesInfo', () => {
       return getHolesInfo(this);
     });
   }
 
-  getExternalIDs(): {
-    externalIDs: number[];
-    externalLengths: number[];
-  } {
+  getExternalIDs(): Border[] {
```
stropitek, line 234: Rename to `getExternalBorders`

src/roi/Roi.ts
```diff
@@ -221,48 +221,48 @@ export class Roi {
...
-        this.#computed.externalIDs.push(borders[i]);
-        this.#computed.externalLengths.push(lengths[i]);
+    for (let border of borders) {
+      if (!internals.includes(border.connectedID)) {
+        const element: Border = {
+          connectedID: border.connectedID,
+          length: border.length,
+        };
+
+        this.#computed.externalBorders.push(element);
```
stropitek, line 250: We created the `getComputed` method so that we don't have to set properties on it ourselves.

Never set or get `#computed` except for in the `getComputed` method.

src/roi/Roi.ts
```diff
@@ -351,7 +353,13 @@ export class Roi {
       };
     });
   }
-
+  /**
+   * A helper function to cache already calculated properties
+   *
+   * @param property
+   * @param callback
+   * @returns Computed[T]
```
stropitek, line 361:
```suggestion
   * A helper function to cache already calculated properties
```

src/roi/Roi.ts
```diff
@@ -364,6 +372,13 @@ export class Roi {
     return this.#computed[property] as Computed[T];
   }
   //TODO Make this private
+  /**
+   * Calculates the correct index on the map of ROI
+   *
+   * @param y
+   * @param x
+   * @returns number
```
stropitek, line 380:
```suggestion
   * Calculates the correct index on the map of ROI
```

src/roi/Roi.ts
```diff
@@ -390,25 +405,37 @@ function getPerimeterInfo(roi: Roi) {
...
+          roi.externalBorders.find((el) => el.connectedID === data[target - 1])
+        ) {
           nbAround++;
         }
 
         if (column === roiMap.width - 1) {
           nbAround++;
-        } else if (roi.externalIDs.includes(data[target + 1])) {
+        } else if (
+          roi.externalBorders.find((el) => el.connectedID === data[target + 1])
```
stropitek, line 417: Can you revert to not using `includes`?

You can recreate the `externalIDs` array before the loop.

src/roi/Roi.ts
```diff
@@ -495,7 +527,12 @@ function getInternalIDs(roi: Roi) {
 
   return internal;
 }
-
+/**
+ * Not sure what it does
+ *
+ * @param roi
+ * @returns
+ */
```
stropitek, line 535: Avoid adding personal comments to the the code

You don't need to add comment to every function.

src/roi/__tests__/borderLengths.test.ts
```diff
@@ -10,22 +10,30 @@ test('border lengths property 5x5', () => {
...
+
+//   const roiMapManager = fromMask(mask);
+
+//   const rois = roiMapManager.getRois({ kind: RoiKind.BW });
+
+//   console.log(rois.map((roi) => roi.borders));
+//   console.log(rois[0].getMap());
+//   expect(rois[0].borders[0]).toStrictEqual({ connectedID: -1, length: 6 });
+// });
+
```
stropitek, line 36: ?

src/roi/Roi.ts
EscapedGibbon: Not sure I understand. What do you mean by "revert to not using 'includes'"? It is not used anymore. Or you want me to revert it to use it again?

src/roi/__tests__/borderLengths.test.ts
EscapedGibbon: It's the test that we talked about before last week. The way borders are calculated for black and white ROI is kind of weird and we were not sure if the behaviour was correct. I think we decided on commenting it to pass the pull request, then uncomment it and create an issue once everything is merged.

targos: If you want to keep a failing test, use `test.failing(...)`.  https://jestjs.io/docs/api#testskipfailingname-fn-timeout

src/roi/Roi.ts
```diff
@@ -245,12 +245,15 @@ export class Roi {
...
         externalBorders.push(element);
       }
     }
-    const externalIDs = externalBorders;
-    return externalIDs;
+
+    return {
+      externalIDs,
+      externalBorders,
+    };
```
stropitek, line 256: You are returning the IDs twice. Please only return `externalBorders`

src/roi/Roi.ts
```diff
@@ -382,33 +393,33 @@ function getPerimeterInfo(roi: Roi) {
   let two = 0;
   let three = 0;
   let four = 0;
-
+  let externalIDs = roi.getExternalBorders().externalIDs;
```
stropitek, line 396: You should use the getter here so that it gets the cached value.

stropitek (approved):

PR #277, opened 2023-04-06
253 new property on roi centroid

EscapedGibbon:


src/roi/Roi.ts
```diff
@@ -328,6 +329,34 @@ export class Roi {
...
+      for (let column = 0; column < this.width; column++) {
+        for (let row = 0; row < this.height; row++) {
+          let target = this.computeIndex(row, column);
+          if (data[target] === this.id) {
+            xCoords.push(column + this.origin.column);
+            yCoords.push(row + this.origin.row);
+          }
+        }
+      }
+      let Xsum = xCoords.reduce((a, b) => {
```
lpatiny, line 347: variables must start with a lowercase ! You may use here xSum

lpatiny: You not nearly never use reduce. It is very slow. You can use directly use xMean from the package 'ml-spectra-processing'

src/roi/Roi.ts
```diff
@@ -328,6 +329,34 @@ export class Roi {
     };
   }
 
+  get centroid() {
+    return this.#getComputed('centroid', () => {
+      const roiMap = this.getMap();
+      const data = roiMap.data;
+      let xCoords = [];
```
lpatiny, line 336: We have here a special case. We know the length of the array at the beginning, it has to be the surface. Moreover in this case we know that it contains only integer. We should therefore use typed array Uint32. It will be in anyway much faster that using 'push'. https://jsbench.me/z8laj0dyuv

lpatiny (changes requested):
I added some comments on your code but in reality the full algorithm has to be redone because:
- it seems not required to create an array, just directly do the sum
- you can first calculate the mean and then add the origin coordinate

Creating the array would however be required if we wanted the median.

src/roi/Roi.ts
```diff
@@ -328,6 +329,29 @@ export class Roi {
...
+    return this.#getComputed('centroid', () => {
+      const roiMap = this.getMap();
+      const data = roiMap.data;
+      let xCoords = 0;
+      let yCoords = 0;
+      for (let column = 0; column < this.width; column++) {
+        for (let row = 0; row < this.height; row++) {
+          let target = this.computeIndex(row, column);
+          if (data[target] === this.id) {
+            xCoords += column + this.origin.column;
```
lpatiny, line 342: The code is better but adding the origin can be done only once when returning the value

PR #286, opened 2023-04-13
test: fromMask of image containing only '1' or '0'

lpatiny:
close: https://github.com/image-js/image-js-typescript/issues/287


targos:
I don't understand. Tests are not failing

Issue #288, opened 2023-04-13
Handle threshold from 0 to 1


opatiny:
Allow to pass either a value from 1 to `maxValue` or in [0,1[ and then handle it as a percentage

opatiny:
In the end we're making a breaking change and only keep the possibility of passing a value from 0 to 1.

PR #290, opened 2023-04-13
fix: handle 0 to 1 threshold

opatiny:
Closes: #288

### DIFF
--- a/src/operations/threshold.ts
+++ b/src/operations/threshold.ts
@@ -1,7 +1,6 @@
 import { Mask } from '..';
 import { Image } from '../Image';
 import { imageToOutputMask } from '../utils/getOutputImage';
-import { validateValue } from '../utils/validators';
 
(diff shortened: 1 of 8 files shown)


src/operations/threshold.ts
```diff
@@ -131,29 +131,16 @@ export function computeThreshold(
...
-      if (percents < 0 || percents > 100) {
-        throw new RangeError(
-          'threshold: threshold in percents is out of range 0 to 100',
-        );
-      }
-      thresholdValue = (percents / 100) * image.maxValue;
-    } else {
-      throw new Error('threshold: unrecognised threshold format');
+    if (threshold < 0 || threshold > 1) {
+      throw new Error('threshold should be a value between 0 and 1');
```
targos, line 138:
```suggestion
      throw new Error('threshold must be a value between 0 and 1');
```

Issue #252, opened 2023-03-25
getMask options and behaviour


lpatiny:
I think allowCorners is not used

https://github.com/image-js/image-js-typescript/blob/c85e82d5c830c334ec0d9f8e4ecb91964f9f8944/src/roi/getMask.ts#L12-L17

The name of the option `innerBorders` does not speak to me. I would rename it to `solidFill` (the name of the method you apply).

https://github.com/image-js/image-js-typescript/blob/c85e82d5c830c334ec0d9f8e4ecb91964f9f8944/src/roi/getMask.ts#L6-L11

PR #292, opened 2023-04-13
fix(getMask): remove useless option

opatiny:
Closes: getMask options and behaviour #252


src/roi/getMask.ts
```diff
@@ -4,27 +4,22 @@ import { Roi } from './Roi';
...
- * Generate a mask of an ROI. You can specify the kind of mask you want using the `kind` option.
+ * Generate a mask of an ROI.
  *
  * @param roi - The ROI to generate a mask for.
  * @param options - Get mask options.
  * @returns The ROI mask.
  */
 export function getMask(roi: Roi, options: GetMaskOptions = {}): Mask {
-  const { innerBorders = true } = options;
+  const { solidFill = true } = options;
```
lpatiny, line 22: Should be false by default I think. And please check that both case are covered by the test cases.

.gitignore
```diff
@@ -130,3 +130,4 @@ src/**/*.png
 
 # ignore images in test scripts folder
 scripts/**.png
+scripts/background.jpg
```
targos, line 133:
```suggestion
scripts/**.jpg
```

targos (approved):

Issue #256, opened 2023-03-26
channelLabels


lpatiny:
Would be nice to have the `channelLabels` for an image.

For a RGBA image this method would return `["Red", "Green", "Blue", "Alpha"]`. This would allow to easily add titles in histograms for example

![image](https://user-images.githubusercontent.com/1484241/227790011-5772a1e8-d936-4c30-8f5f-c08117ced913.png)

PR #294, opened 2023-04-16
feat: implement channelLabels

opatiny:
Closes: #256

### DIFF
--- a/src/utils/constants/channelLabels.ts
+++ b/src/utils/constants/channelLabels.ts
@@ -1,7 +1,9 @@
+import { ImageColorModel } from './colorModels';
+
 export const channelLabels = {
   GREY: ['Grey'],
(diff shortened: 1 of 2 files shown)


src/utils/constants/channelLabels.ts
```diff
@@ -0,0 +1,7 @@
+export const channelLabels = {
+  GREY: ['Grey'],
+  GREYA: ['Grey', 'Alpha'],
+  RGB: ['Red', 'Green', 'Blue'],
+  RGBA: ['Red', 'Green', 'Blue', 'Alpha'],
+  BINARY: ['Mask'],
+} as const;
```
targos, line 7:
```suggestion
} as const satisfies Record<ImageColorModel, string[]>;
```

This should make sure we don't forget to update the object if we add new color models.

PR #296, opened 2023-04-16
perf: optimize bilinear

lpatiny:
This PR improve the resize speed by a factor 2 to 4.

- improve bilinear
- improve round


src/utils/round.ts
```diff
@@ -5,8 +5,13 @@
  * @returns The rounded value.
  */
 export function round(value: number): number {
-  if (value % 0.5 !== 0) {
-    return Math.round(value);
+  const integer = value >> 0;
```
targos, line 8: What about `Math.trunc(value)` ? To avoid unnecessary conversion to a 32-bit integer.

lpatiny: You know I like bit shifts ... but I'm wrong !
Using `Math.trunc` is 2 times faster !!!!

With my bit shift: 10.719s, with Math.trunc: 5.952s. Clamped array:  5.302s

PR #308, opened 2023-04-18
279 add new method getroibyid on roimapmanager

EscapedGibbon:


targos:
In general, please take the time to write a pull request description in English (do not use the default one based on the branch name)

src/roi/RoiMapManager.ts
```diff
@@ -66,6 +66,15 @@ export class RoiMapManager implements RoiManager {
     return this.map.data[this.map.width * row + column];
   }
 
+  public getRoiById(roiID: number): Roi {
+    const rois = this.getRois();
+    const thatRoi = rois.find((roi) => roi.id === roiID);
```
lpatiny, line 71: Could be renamed to `foundRoi`

src/roi/RoiMapManager.ts
```diff
@@ -66,6 +66,15 @@ export class RoiMapManager implements RoiManager {
...
+    if (thatRoi === undefined) {
+      throw new Error(`invalid ID: ${roiID}`);
+    } else {
```
lpatiny, line 74: The else is not necessary

src/roi/RoiMapManager.ts
```diff
@@ -66,6 +66,15 @@ export class RoiMapManager implements RoiManager {
     return this.map.data[this.map.width * row + column];
   }
 
+  public getRoiById(roiID: number): Roi {
+    const rois = this.getRois();
+    const thatRoi = rois.find((roi) => roi.id === roiID);
+    if (thatRoi === undefined) {
```
lpatiny, line 72: In this case I would write
```
if(! thatRoi) {

}
```

lpatiny (changes requested):

PR #309, opened 2023-04-19
246 new roi property ellipse

EscapedGibbon:


src/roi/Roi.ts
```diff
@@ -608,3 +641,105 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
+    ) *
+    Math.PI;
+  return {
+    rMajor,
+    rMinor,
+    position: {
+      x: xCenter,
+      y: yCenter,
+    },
+    majorAxis: {
```
lpatiny, line 735: Could you use the same kind of object as FeretDiameter
```
export interface FeretDiameter {
  /**
   * Start and end point of the Feret diameter.
   */
  points: Point[];
  /**
   * Length of the diameter.
   */
  length: number;
  /**
   * Angle between the diameter and a horizontal line in degrees.
   */
  angle: number;
}
```

src/roi/Roi.ts
```diff
@@ -608,3 +641,105 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
+    Math.sqrt(
+      Math.pow(minorAxisPoint1.x - xCenter, 2) +
+        Math.pow(minorAxisPoint1.y - yCenter, 2),
+    ) *
+    Math.PI;
+  return {
+    rMajor,
+    rMinor,
+    position: {
+      x: xCenter,
```
lpatiny, line 732: In this project we should use column and row as variable names. You could replace everywhere x and y by column and row (in all variable names)

src/roi/Roi.ts
```diff
@@ -608,3 +641,105 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
+  //spectral decomposition of the sample covariance matrix
+  let sampleCovarianceMatrix = [
+    [centeredXVariance, centeredCovariance],
+    [centeredCovariance, centeredYVariance],
+  ];
+  let e = new EigenvalueDecomposition(sampleCovarianceMatrix);
+  let eigenvalues = e.realEigenvalues;
+  let vectors = e.eigenvectorMatrix;
+
+  let rMajor: number;
```
lpatiny, line 676: Please rename to radiusMajor and radiusMinor (I guess those are radius)

src/roi/Roi.ts
```diff
@@ -608,3 +641,105 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
...
+      Math.pow(majorAxisPoint1.x - xCenter, 2) +
+        Math.pow(majorAxisPoint1.y - yCenter, 2),
+    ) *
+    Math.sqrt(
+      Math.pow(minorAxisPoint1.x - xCenter, 2) +
+        Math.pow(minorAxisPoint1.y - yCenter, 2),
+    ) *
+    Math.PI;
+  return {
+    rMajor,
```
lpatiny, line 729: We don't need rMajor and rMinor we will have both length in the minorAxis and majorAxis properties

package.json
```diff
@@ -49,7 +49,10 @@
     "image-type": "^4.1.0",
     "jpeg-js": "^0.4.4",
     "median-quickselect": "^1.0.1",
+    "ml-array-variance": "^1.1.8",
```
lpatiny, line 52: Please check if we don't have this method in ml-spectra-processing so that you don't need to add one more package.

Same is true for ml-array-xy-covariance

src/roi/Roi.ts
```diff
@@ -13,6 +18,23 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 
+interface Ellipse {
+  rMajor: number;
+  rMinor: number;
+  position: {
```
lpatiny, line 24: I think we could rename `position` to `center`

src/roi/Roi.ts
```diff
@@ -608,3 +637,101 @@ function getBorders(roi: Roi): { ids: number[]; lengths: number[] } {
     lengths: borderLengths,
   };
 }
+
+function getEllipse(
+  roi: Roi,
+  options: { nbSD: number; scale: number },
```
lpatiny, line 643: Please remove those options. The expected current behaviour is to return the ellipse with the same surface.

lpatiny (approved):

package.json
lpatiny: https://github.com/mljs/spectra-processing/blob/0e0361bab81e8a12c700b4429e468f92182370ad/src/x/xVariance.ts#L12-L14 can not be used ?

src/roi/Roi.ts
```diff
@@ -626,3 +654,103 @@ function getBorders(roi: Roi): Border[] {
...
+  let e = new EigenvalueDecomposition(sampleCovarianceMatrix);
+  let eigenvalues = e.realEigenvalues;
+  let vectors = e.eigenvectorMatrix;
+
+  let radiusMajor: number;
+  let radiusMinor: number;
+  let vectorMajor: number[];
+  let vectorMinor: number[];
+
+  if (eigenvalues[0] > eigenvalues[1]) {
```
targos, line 692: Is it possible to add a test that covers this condition?

src/roi/Roi.ts
```diff
@@ -626,3 +654,103 @@ function getBorders(roi: Roi): Border[] {
     };
   });
 }
+
+function getEllipse(roi: Roi, scale: number): Ellipse {
```
targos, line 658: Please move this function to another file (in a subfolder: `./properties/getEllipse`. Roi.ts is becoming really too big.

src/roi/Roi.ts
EscapedGibbon: > Is it possible to add a test that covers this condition?

I don't think it is possible to create a test that covers it because the eigenvalue array is sorted at the end of its computation. So it is impossible for element 0 to be bigger than element 1.

targos: If you're confident that the condition is impossible, you can remove it and add `assert(eigenvalues[0] <= eigenvalues[1]);` instead (be careful to import assert from https://github.com/image-js/image-js-typescript/blob/main/src/utils/assert.ts.

lpatiny:
@EscapedGibbon Please add the following test case as well

0 1 0
0 1 0
0 1 0

Currently it yields to NaN. Could be improved in the future.

src/roi/Roi.ts
```diff
@@ -392,6 +394,17 @@ export class Roi {
     });
   }
 
+  get ellipse(): Ellipse {
+    return this.#getComputed('ellipse', () => {
+      let ellipse = getEllipse(this, 1);
+      if (ellipse.surface !== this.surface) {
+        const scaleFactor = Math.sqrt(this.surface / ellipse.surface);
+        ellipse = getEllipse(this, scaleFactor);
```
lpatiny, line 402: I would have never done it like that because this process (calculate the ellipse) is CPU intensive. Please use as second parameter the surface `getEllipse(this, surface)` and `getEllipse` can calculate the scaleFactor internally.

targos:
I can have a look, but in the end it's @lpatiny who knows best about this feature

PR #312, opened 2023-04-20
Handle arrays in `level` options

opatiny:
- feat: expose image.minMax
- feat(level): parameters can now be arrays
- test: enhance level coverage
- feat: implement enhanceContrast
- feat(drawRectangle): add fillColor = 'none' option


src/filters/enhanceContrast.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Enhance the contrast of an image by spanning each channel on the range [0, image.maxValue].
+ *
+ * @param image - The image to enhance.
+ * @param options - Enhance contrast options.
+ * @returns The enhanced image.
+ */
+export function enhanceContrast(
```
targos, line 20: I'm not sure about the name. This is one way of enhancing contrast, but there are others and this one doesn't deserve such a generic name more, does it?

src/filters/level.ts
```diff
@@ -94,3 +108,25 @@ export function level(image: Image, options: LevelOptions = {}) {
...
+ */
+function getValueArray(
+  value: number | number[],
+  imageChannels: number,
+): number[] {
+  if (Array.isArray(value)) {
+    if (value.length === imageChannels) {
+      return value;
+    } else {
+      throw new Error('array length is not compatible with channel option');
```
targos, line 127:
```suggestion
      throw new RangeError('array length is not compatible with channel option');
```

src/filters/enhanceContrast.ts
opatiny: Indeed, do you have any other idea?

opatiny: Proposition: `spanChannelToFullRange()`

targos: what about `autoLevel` ?

opatiny: sgtm!

src/filters/autoLevel.ts
```diff
@@ -0,0 +1,47 @@
...
+    .fill(0)
+    .map((value, index) => index);
+
+  if (image.colorModel === ImageColorModel.GREYA) {
+    channels = [0];
+  } else if (image.colorModel === ImageColorModel.RGBA) {
+    channels = [0, 1, 2];
+  }
+
+  return newImage.level({
```
targos, line 39: can't we use `image.level` directly without creating a `newImage` in this function?

targos (approved):

PR #317, opened 2023-04-21
test: replace Jest with Vitest

targos:
Highlights:
- Update all dependencies.
- Use Vitest instead of Jest and adapt config.
- Make the default package type "module".
  - A CommonJS build is still done, but is now the secondary one.
- Update tsconfig to be more modern and ESM-first.
- Avoid circular dependencies by importing from deeper than `src/index.ts` internally.
- Use `ts-pattern` instead of `switch` statements.


targos:
I'll wait for https://github.com/image-js/image-js-typescript/pull/480 to avoid conflicts in it.

PR #318, opened 2023-04-22
315 move helper functions into getters

EscapedGibbon:


stropitek (approved):

targos:
We obviously have a different opinion here, but I would prefer to keep separate functions and move them to other files to avoid a single class with hundreds of lines.

stropitek:
> We obviously have a different opinion here, but I would prefer to keep separate functions and move them to other files to avoid a single class with hundreds of lines.

Sorry I just noticed the comment after merging 😨

stropitek:
Feel free to open another issue to split by files.

What I wished for with this was to allow some roi properties to become private (like the `map` property)

This means if we create helpers in other files, they have to take those private properties as argument, they cannot just work with the Roi instance.

PR #337, opened 2023-05-12
feat: implement median filter for noise removal

EscapedGibbon:


src/filters/medianFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+    kernel = new Uint16Array(size);
+  } else {
+    kernel = new Uint8Array(size);
+  }
+
+  for (let channel = 0; channel < image.channels; channel++) {
+    for (let row = 0; row < image.height; row++) {
+      for (let column = 0; column < image.width; column++) {
+        let n = 0;
+        for (let cellSizeRow = -kSize; cellSizeRow <= kSize; cellSizeRow++) {
```
lpatiny, line 63: cellRow and cellColumn would be better names I think

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+  checkProcessable(image, {
+    bitDepth: [8, 16],
+  });
+
+  if (cellSize < 1) {
+    throw new RangeError('cellSize must be greater than 0');
+  }
+
+  if (cellSize % 2 === 0) {
+    throw new Error('cellSize must be an odd number');
```
lpatiny, line 44: @targos If this also a RangeError ?

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  // validateChannels(options.channels as number[], image);
+  let interpolateBorder = getBorderInterpolation(borderType, borderValue);
+  let kSize = cellSize;
+  let newImage = Image.createFrom(image);
+  let size = (kSize * 2 + 1) * (kSize * 2 + 1);
+  let kernel;
```
lpatiny, line 52: Just define it always as new Uint16Array(size). It may always be problematic that the variable may be 2 different kinds.

lpatiny: `kernel` seems a strange name. Maybe `cellValues`

src/Image.ts
```diff
@@ -941,6 +993,20 @@ export class Image {
   ): Image {
     return correctColor(this, measuredColors, referenceColors);
   }
+  /**
+   * Apply a median filter to the image.
+   *
+   * @param options - Options to apply for median filter.
+   * @param options.radius - Size of the area to calculate median value from.
```
lpatiny, line 1000: You should only keep options, all the other parameters are defined in MedianFilterOptions

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Value of border.
+   */
+  borderValue: number;
+  /**
+   * The radius of the cell to extract xMedian from. Must be odd.
+   *
+   * @default 1
+   */
+  cellSize: number;
```
lpatiny, line 23: If we call it `cellSize` (like in pixelate) it should be be a radius but the actual size.

lpatiny (changes requested):

src/filters/medianFilter.ts
targos: yes

EscapedGibbon:
Conversations resolved

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  if (cellSize < 1) {
+    throw new RangeError('cellSize must be greater than 0');
+  }
+
+  if (cellSize % 2 === 0) {
+    throw new RangeError('cellSize must be an odd number');
+  }
+
+  // validateChannels(options.channels as number[], image);
```
stropitek, line 47: ?

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @default 'reflect101'
+   */
+  borderType: BorderType;
+  /**
+   * Value of border.
+   */
+  borderValue: number;
+  /**
+   * The radius of the cell to extract xMedian from. Must be odd.
```
stropitek, line 19: "xMedian" is confusing (I suppose it's related to implementation and should not be in the documentation)

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @default 1
+   */
+  cellSize: number;
+}
+/**
+ * Apply a median filter to the image.
```
stropitek, line 26: Please don't paraphrase the function's name. It should describe what the function does in 2-3 sentences, not state what the function is.

src/filters/medianFilter.ts
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Type of border algorithm to interpolate from.
+   *
+   * @default 'reflect101'
+   */
+  borderType: BorderType;
+  /**
+   * Value of border.
+   */
+  borderValue: number;
```
stropitek, line 17: In other methods of image-js having the `borderType` option, is this option also a required field?

EscapedGibbon:
Done

src/filters/medianFilter.ts
```diff
@@ -14,16 +14,16 @@ export interface MedianFilterOptions {
...
-   * The radius of the cell to extract xMedian from. Must be odd.
+   * The radius of the cell to extract median value from. Must be odd.
    *
    * @default 1
    */
   cellSize: number;
 }
 /**
- * Apply a median filter to the image.
+ * Calculate a new image that replaces all pixel values by the median value inside the square area set in the options
```
stropitek, line 26:
```suggestion
 * Calculate a new image that replaces all pixel values by the median of neighbouring pixels.
```

PR #341, opened 2023-05-22
fix: calculate aspectRatio more consistently 

EscapedGibbon:


src/maskAnalysis/utils/getMbrFromPoints.ts
```diff
@@ -91,10 +91,16 @@ export function getMbrFromPoints(points: readonly Point[]): Mbr {
   }
 
   const mbrRotated = rotate(minSurfaceAngle, mbr);
-  const width = mbr[0].column - mbr[2].column;
-  const height = mbr[0].row - mbr[2].row;
+  const width = Math.max(
```
lpatiny, line 94: Please take care to variable names. This is not the width but the maxSide (or a better name ?).

src/maskAnalysis/utils/getMbrFromPoints.ts
```diff
@@ -91,10 +91,16 @@ export function getMbrFromPoints(points: readonly Point[]): Mbr {
...
+    mbr[0].column - mbr[2].column,
+    mbr[0].row - mbr[2].row,
+  );
+  const height = Math.min(
```
lpatiny, line 98: Same name problem

src/maskAnalysis/utils/getMbrFromPoints.ts
```diff
@@ -91,10 +91,16 @@ export function getMbrFromPoints(points: readonly Point[]): Mbr {
...
+    mbr[0].column - mbr[2].column,
+    mbr[0].row - mbr[2].row,
+  );
   const mbrAngle = getMbrAngle(mbrRotated);
-  const ratio = height / width;
+  const ratio = Math.abs(height / width);
```
lpatiny, line 103: The name could be already `aspectRatio`.

Math.abs is a problem for me because if we are comparing before min / max and one of the argument may be negative we are in trouble.

We need then to calculate before `width`as Math.abs(...), `height` as Math.abs() and use those values to calculate `minSide` and `maxSide`

EscapedGibbon: Understood. We could calculate width and height through the norm of vectors that the points make, something like Math.sqrt((x1-x2)**2 +(y1+y2)**2). Then just put the one which is bigger as width. Or is it too consuming performance-wise?

lpatiny: There is a specific function that allows to do this: `Math.hypot`

Issue #316, opened 2023-04-21
Cleanup comments in Roi.ts


stropitek:
There are a lot of eslint warnings like this one

![CleanShot 2023-04-21 at 10 33 30](https://user-images.githubusercontent.com/4118690/233587587-053990a6-75db-4cbd-862b-04cf0886f446.png)

Please fix.

Remove comments which don't add any context or information like this one:

![CleanShot 2023-04-21 at 10 34 18](https://user-images.githubusercontent.com/4118690/233587803-4e3ff415-aa97-49c2-8210-6b6c7814cd23.png)

targos:
We should not remove any comments but fix them with proper documentation.

PR #342, opened 2023-05-22
docs: improve roi comments

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/316

### DIFF
--- a/src/roi/Roi.ts
+++ b/src/roi/Roi.ts
@@ -16,11 +16,15 @@ import { Ellipse, getEllipse } from './properties/getEllipse';
 /**
  * Properties of borders of ROI.
  *
- * connectedID - Refers  to the roiID of the contiguous ROI.
- * number - Length of the border with connectedID.
  */
 interface Border {
+  /**
+   * Refers  to the roiID of the contiguous ROI.
+   */
   connectedID: number;
+  /**
+   * Length of the border with connectedID.
+   */
   length: number;
 }
 interface Computed {
@@ -590,7 +594,7 @@ export class Roi {
   /**
    * A JSON object with all the data about ROI.
    *
-   * @returns Calculated properties as one object.
+   * @returns All current ROI properties as one object.
    */
   toJSON() {
     return {
--- a/src/roi/Roi.ts
+++ b/src/roi/Roi.ts
@@ -14,15 +14,15 @@ import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 import { Ellipse, getEllipse } from './properties/getEllipse';
 /**
- * Interface for borders of ROI.
+ * Properties of borders of ROI.
+ *
+ * @param connectedID - Refers  to the roiID of the contiguous ROI.
+ * @param number - Length of the border with connectedID.
  */
 interface Border {
-  connectedID: number; // refers to the roiID of the contiguous ROI
+  connectedID: number;
   length: number;
 }
-/**
- * Interface with cached values of computed properties.
- */
 interface Computed {
   perimeter: number;
   borders: Border[]; // external and internal ids which are not equal to the current roi ID
@@ -125,7 +125,7 @@ export class Roi {
   /**
    * Computes the diameter of a circle that has the same perimeter as the particle image.
    *
-   * @returns ped value in pixels
+   * @returns Ped value in pixels.
    */
   get ped() {
     return this.perimeter / Math.PI;
@@ -146,7 +146,7 @@ export class Roi {
    * Returns an array of ROIs IDs that are included in the current ROI.
    * This will be useful to know if there are some holes in the ROI.
    *
-   * @returns internalIDs
+   * @returns InternalIDs.
    */
   get internalIDs() {
     return this.#getComputed('internalIDs', () => {
@@ -171,7 +171,7 @@ export class Roi {
         for (let row = 1; row < this.height - 1; row++) {
           let target = this.#computeIndex(row, column);
           if (internal.includes(data[target])) {
-            // we check if one of the neighbour is not yet in
+            // We check if one of the neighbor is not yet in.
 
             array[0] = data[target - 1];
             array[1] = data[target + 1];
@@ -195,11 +195,11 @@ export class Roi {
   /**
    * Returns an array of ROIs IDs that touch the current ROI.
    *
-   * @returns array of Borders
+   * @returns The array of Borders.
    */
   get externalBorders(): Border[] {
     return this.#getComputed('externalBorders', () => {
-      // take all the borders and remove the internal one ...
+      // Takes all the borders and removes the internal one ...
       let borders = this.borders;
 
       let externalBorders = [];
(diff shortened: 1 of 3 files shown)


src/roi/Roi.ts
```diff
@@ -128,6 +128,8 @@ export class Roi {
 
   /**
    * Diameter of a circle of equal perimeter
+   *
+   * @returns ped value
```
stropitek, line 132: Maybe specify the unit: "in pixels"

src/roi/Roi.ts
```diff
@@ -198,7 +200,9 @@ export class Roi {
   //TODO externalIds should be an array of {id: number, length: number}
```
stropitek, line 200: Maybe you can also look at cleaning up TODOs in this PR?

src/roi/Roi.ts
```diff
@@ -198,7 +200,9 @@ export class Roi {
...
 
   /**
-   * Return an array of ROIs IDs that touch the current ROI.
+   * Returns an array of ROIs IDs that touch the current ROI.
+   *
+   * @returns array of IDs
```
stropitek, line 205: The return type `Border` is not documented

src/roi/Roi.ts
```diff
@@ -295,7 +298,10 @@ export class Roi {
    * The perimeter is calculated using the sum of all the external borders of the ROI to which we subtract
    * (2 - √2) * the number of pixels that have 2 external borders
    * 2 * (2 - √2) * the number of pixels that have 3 external borders
+   *
+   * @returns perimeter value
```
stropitek, line 302: unit?

src/roi/Roi.ts
```diff
@@ -387,6 +395,8 @@ export class Roi {
 
   /**
    * Returns the diameter of a circle of equal projection area
+   *
+   * @returns eqpc value
```
stropitek, line 399: unit?

src/roi/Roi.ts
```diff
@@ -387,6 +395,8 @@ export class Roi {
 
   /**
    * Returns the diameter of a circle of equal projection area
```
stropitek, line 397: Prefer "Computes" over "Returns". The fact that it "returns" is already documented.

stropitek (commented):
Can you try to expand a bit the text describing each getter / method?

The idea is to say more than what is in the function's name. Try to explain in simple terms (1-3 sentences).

For example for `solidity`, instead of:

> Calculates solidity of the ROI

You could write something like this (copied + edited from the internet):

> The solidity describes the extent to which a shape is convex or concave. The solidity of a completely convex shape is 1, the farther the solidity deviates from 1, the greater the extent of concavity in the shape of the ROI.

stropitek:
Take care to be consistent with the usage of periods (`.`). Some sentences have them and some don't. Also a period should always be followed by a space before the next sentence. Avoid long sentences in parenthesis.

stropitek:
A sentence should start with an upper case AND have a period at the end.

@targos is adding an eslint rule to automate this. Once it is on main you can merge main into this branch and it will be auto-fixable.

src/roi/Roi.ts
```diff
@@ -13,11 +13,16 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 import { Ellipse, getEllipse } from './properties/getEllipse';
-
+/**
+ * Interface for borders of ROI.
+ */
 interface Border {
   connectedID: number; // refers to the roiID of the contiguous ROI
   length: number;
```
stropitek, line 21: can you document these 2 properties as jsdoc block comments?

src/roi/Roi.ts
```diff
@@ -13,11 +13,16 @@ import { RoiMap } from './RoiMapManager';
...
 }
+/**
+ * Interface with cached values of computed properties.
```
stropitek, line 24: Avoid saying it is an "interface". We already know that by reading the code.

src/roi/Roi.ts
```diff
@@ -13,11 +13,16 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 import { Ellipse, getEllipse } from './properties/getEllipse';
-
+/**
+ * Interface for borders of ROI.
```
stropitek, line 17: ditto for "interface"

Since border is exposed (returned by some public property of ROI), then this interface should have a more specific name (maybe `ROIBorder`) and be exported

src/roi/Roi.ts
```diff
@@ -195,10 +202,11 @@ export class Roi {
       return internal;
     });
   }
-  //TODO externalIds should be an array of {id: number, length: number}
 
   /**
-   * Return an array of ROIs IDs that touch the current ROI.
+   * Returns an array of ROIs IDs that touch the current ROI.
+   *
+   * @returns array of Borders
```
stropitek, line 209: Use sentences everywhere, not just the description.

targos:
@EscapedGibbon Can you please make this a priority?

src/roi/Roi.ts
EscapedGibbon: "Interface" removed from the description. Change of the name of the interface should probably be done on a different branch.

EscapedGibbon:
Done.

src/roi/Roi.ts
```diff
@@ -662,14 +663,14 @@ export class Roi {
...
 
   /**
    * Calculates the correct index on the map of ROI.
    *
-   * @param y - map row
-   * @param x - map column
-   * @returns index of ROI map
+   * @param y - Map row
+   * @param x - Map column
+   * @returns Index of ROI map.
```
stropitek, line 672:
```suggestion
   * @returns Index within the ROI map.
```

src/roi/Roi.ts
```diff
@@ -500,56 +517,90 @@ export class Roi {
...
   get feret(): Feret {
     return this.#getComputed('feret', () => {
       return getFeret(this.getMask());
     });
   }
   /**
+   * A JSON object with all the data about ROI.
    *
-   * @returns calculated properties as one object
+   * @returns Calculated properties as one object.
```
stropitek, line 603: It is more than the calculated properties. I guess it's supposed to be "All ROI properties"

src/roi/Roi.ts
```diff
@@ -13,9 +13,14 @@ import { RoiMap } from './RoiMapManager';
 import { getBorderPoints } from './getBorderPoints';
 import { getMask, GetMaskOptions } from './getMask';
 import { Ellipse, getEllipse } from './properties/getEllipse';
-
+/**
+ * Properties of borders of ROI.
+ *
+ * connectedID - Refers  to the roiID of the contiguous ROI.
+ * number - Length of the border with connectedID.
+ */
```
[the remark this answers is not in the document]
stropitek: The reason I insist is because the documentation should be visible during TS autocomplete. You can check yourself that it is the case.

PR #344, opened 2023-05-24
feat: add option to choose an algorithm of pixelization

EscapedGibbon:


src/filters/pixelate.ts
```diff
@@ -35,32 +41,40 @@ interface CenterOptions {
...
-            column,
-            row,
-          },
+          origin: { column, row },
         });
-
-        const value = image.getValue(center.column, center.row, channel);
-
+        if (algorithm === 'mean') {
+          value = xMean(valuesOfSector);
```
stropitek, line 64: `xMean` is misleading because could indicate the mean is "directional". Maybe just call it `mean`? or `cellMean`?

src/filters/pixelate.ts
```diff
@@ -1,11 +1,21 @@
...
 
 export interface PixelateOptions {
   /**
-   *  range of pixelated area
+   *  range of pixelated area.
    */
   cellSize: number;
+  /**
+   * algorithm to use.
+   */
```
lpatiny, line 17: You should add `@default center` to specify the default value.

src/filters/pixelate.ts
```diff
@@ -1,11 +1,21 @@
+import {
+  xMean as cellMean,
+  xMedian as cellMedian,
```
lpatiny, line 3: You should not rename this method.

src/filters/pixelate.ts
```diff
@@ -49,17 +59,34 @@ export function pixelate(image: Image, options: PixelateOptions): Image {
       for (let row = 0; row < image.height; row += cellSize) {
         const currentCellWidth = Math.min(cellSize, image.width - column);
         const currentCellHeight = Math.min(cellSize, image.height - row);
-
-        const center = getCenter({
+        let valuesOfSector = getArrayOfValues(image, channel, {
```
lpatiny, line 62: To calculate mean you do not need to create a temporary array. It still takes time and memory.

You can directly create 3 (internal) methods:

* getCellCenter()
* getCellMedian()
* getCellMean()

src/filters/pixelate.ts
```diff
@@ -96,3 +123,25 @@ function getCenter(options: CenterOptions): Point {
...
+    for (
+      let row = options.origin.row;
+      row < options.origin.row + options.height;
+      row++
+    ) {
+      array.push(image.getValue(column, row, channel));
+    }
+  }
+  return array;
+}
```
lpatiny, line 147: you should create a new function

```
getValueFunction(algorithm) {
  switch(algorithm) {
    case 'center':
      return getCellCenter; // we return the function !!! This will be the `getValue` function
    ...
  }
}
```
This function should be called before the loop and we can then use `getValue()` inside the loop


To calculate mean you do not need to create a temporary array. It still takes time and memory.

You may also create 3 methods

- getCellCenter()
- getCellMedian()
- getCellMean()

src/filters/pixelate.ts
```diff
@@ -78,13 +83,19 @@ export function pixelate(image: Image, options: PixelateOptions): Image {
 }
 
 /**
- *Find the center of a rectangle to be pixelated
+ *Computes the center value for the current sector
  *
```
lpatiny, line 87: there will soon be a new rule that forbid empty lines between title and options

Issue #340, opened 2023-05-21
getRow / getColumn


lpatiny:
In order to determine possible baseline problems or right level between background and ROI it is important to be able to plot a row and a column:

![image](https://github.com/image-js/image-js-typescript/assets/1484241/b94597a4-ad7a-4c90-84bd-5d099e02ba5a)

Like `getPixel`

https://github.com/image-js/image-js-typescript/blob/258c5e73ec8f5a844326116044ab854bf1443915/src/Image.ts#L308-L322

We would need 2 new methods: `getRow(row: number)` and `getColumn(column: number)` that will return an array of array. The first level of the array is the channel and the second is all the values for this channel.

PR #346, opened 2023-05-26
feat: add getRow and getColumn methods to Image.ts

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/340


stropitek:
Can you add a test with a grey image with more than 1 row and 1 column?

EscapedGibbon:
> Can you add a test with a grey image with more than 1 row and 1 column?

Done

Issue #272, opened 2023-04-03
Code over complex ?


lpatiny:
Is it really necessary to use a call back to call a simple method ?

https://github.com/image-js/image-js-typescript/blob/04f73e6f3f6cf700975760c34707aa07442f37c1/src/roi/Roi.ts#L125-L129

https://github.com/image-js/image-js-typescript/blob/04f73e6f3f6cf700975760c34707aa07442f37c1/src/roi/Roi.ts#L331-L341

For me it makes the code very complex to read and to maintain (and also to contribute in fact).

targos:
I agree. We should only use `#getComputed` when the computation has a big performance cost.

stropitek:
TODO:
remove `#getComputed` from `ped` and `eqpc`

PR #352, opened 2023-05-31
chore: simplify ROI code 

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/272


stropitek:
@EscapedGibbon 

Just realized after merging: you did not remove those 2 properties from the type of the computed properties.

PR #354, opened 2023-06-06
feat: decode and expose metadata

Xstoudi:
Metadata type is directly taken from `image-js`, notable exception is the `gps` field that was removed as I didn't find any use of it in the old version and wasn't tested. I didn't find an example where it was filled with data.

Tell me if you have recommendation or want changes.


src/Image.ts
```diff
@@ -103,6 +103,7 @@ import {
   threshold,
   ThresholdOptions,
 } from './index';
+import { ImageMetadata } from './load/getMetadata';
```
targos, line 106: Can you fix the ESLint warnings here?

src/load/decodeJpeg.ts
```diff
@@ -1,6 +1,8 @@
 import { decode } from 'jpeg-js';
+import { decode as decodeExif } from 'fast-jpeg';
```
targos, line 2: ditto

src/load/decodeTiff.ts
```diff
@@ -41,7 +42,7 @@ function getImageFromIFD(ifd: TiffIfd): Image {
       // TODO: handle other bit depths
       bitDepth: 16,
       // TODO: implement metadata
```
targos, line 44: I guess we can remove this todo

targos (approved):

Xstoudi:
I failed a rebase and it generated bad commits but it should pass now.

Issue #359, opened 2023-06-16
add and improve computeMinima/computeMaxima functions 


EscapedGibbon:
The idea is to add a function that calculates the minima and maxima points, so that later a watershed filter could be integrated. Is it okay to add this function in the compute folder?

https://github.com/image-js/image-js/blob/d0b4e00f37ea01510cb9b4422c95d5285f3c8bab/src/image/compute/localMaxima.js#L14C1-L26

PR #369, opened 2023-06-23
feat: add function to compute minima and maxima 

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/359


src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,61 @@
...
+  const isMax = kind === 'maximum';
+
+  let sortedPoints = points
+    .sort((a, b) => {
+      return Math.min(
+        image.getValue(a.column, a.row, 0),
+        image.getValue(b.column, b.row, 0),
+      );
+    })
+    .reverse();
```
lpatiny, line 36: Just make a conditional reverse depending if it is min or max. This will avoid to have to add a condition in the loop.

src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (removeClosePoints > 0) {
+    for (let i = 0; i < sortedPoints.length; i++) {
+      for (let j = i + 1; j < sortedPoints.length; j++) {
+        if (
+          Math.hypot(
+            sortedPoints[i].column - sortedPoints[j].column,
+            sortedPoints[i].row - points[j].row,
```
lpatiny, line 44: Why do you have sometimes points and sometimes sortedPoints ?

src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function filterPoints(
+  points: Point[],
+  image: Image,
+  options: FilterPointsOptions,
+) {
+  let { removeClosePoints = 0, kind = 'maximum' } = options;
+  const isMax = kind === 'maximum';
+
+  let sortedPoints = points
```
lpatiny, line 29: .sort is an inplace operation. You can not do this because this method will actually change the order of the points and we don't want a function has side effects.

So you should first 'slice' the original array.

src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,61 @@
...
+    .sort((a, b) => {
+      return Math.min(
```
lpatiny, line 31: This does not make sense. A sort must return a negative, positive or zero value and in this case it would always be >=0. It should be a difference.
I'm surprised your testcases didn't detect this issue.

src/compute/getExtrema.ts
```diff
@@ -0,0 +1,107 @@
...
+   * Uses mask to check if a point belongs to a ROI or not
+   */
+  mask?: Mask;
+  /**
+   * Chooses what kind of coverage algorithm to use to compute the extremum.
+   * @default 'star'
+   */
+  algorithm?: 'cross' | 'square' | 'star';
+  /**
+   * Maximum number of points that can be equal to the extremum
```
lpatiny, line 21: What is the default value ?

lpatiny:
I don't think the points filtering is working. When you design testcases you need to think about what can fails.

How does it happens that you remove some files in this PR that is not related to your work ?

lpatiny (changes requested):

src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @returns Array of Points.
+ */
+export function filterPoints(
+  points: Point[],
+  image: Image,
+  options: FilterPointsOptions,
+) {
+  let { removeClosePoints = 0, kind = 'maximum' } = options;
+  const isMax = kind === 'maximum';
+
```
lpatiny, line 27: You are using the first channel of the image and this is quite arbitrary.
Please allow an option 'channel'. If channel is unspecified and nb channels of image > 1, throw an error.

src/utils/geometry/filterPoints.ts
```diff
@@ -0,0 +1,56 @@
...
+  let sortedPoints = points.slice().sort((a, b) => {
+    return (
+      image.getValue(a.column, a.row, 0) - image.getValue(b.column, b.row, 0)
+    );
+  });
```
lpatiny, line 32: Please reverse the array rather than having isMax in the loop.

lpatiny (changes requested):

PR #379, opened 2023-07-31
feat: add watershed filter

EscapedGibbon:


src/filters/waterShed.ts
```diff
@@ -0,0 +1,196 @@
...
+      mask,
+    });
+    points = filterPoints(points, image, { kind });
+    for (let i = 0; i < points.length; i++) {
+      switch (isMinimum) {
+        case true:
+          if (
+            image.getValue(points[i].column, points[i].row, 0) >= fillMaxValue
+          ) {
+            points.splice(i, 1);
```
lpatiny, line 104: You should use points.filter( () => {})

src/filters/waterShed.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+    result.push(rowValues);
+  }
+  let resultMask = createMask(result);
+  if (isMinimum) {
+    resultMask = resultMask.invert();
+  }
+  //console.log(resultMask.invert(), fromMask(resultMask).getRois());
+
+  return fromMask(resultMask);
```
lpatiny, line 195: We need to return a RoiMapManager seems to me. What is a mask doing here ?

EscapedGibbon: It does return a RoiMapManager. The problem that I encountered was that without nbPositive and nbNegative the RoiMapManager will not be correct. The only certain way I found to calculate it is to create a mask which then could be transformed into a RoiMapManager with correct number of positive and negative values on the map

lpatiny: But you should not go through a Mask. You need to directly create:

https://github.com/image-js/image-js-typescript/blob/24a6b6622ead70c2264322501d9cf95279733554/src/roi/RoiMapManager.ts#L43-L47

A code similar to:

https://github.com/image-js/image-js-typescript/blob/24a6b6622ead70c2264322501d9cf95279733554/src/roi/fromMask.ts#L176-L182

EscapedGibbon: I understand but I need to know the nbNegative and nbPositive need to give the number of positive and negative Rois. And I don't know how to compute those without a createMask function. For example, in this example there are 2 negative rois and 2 positive ones. And if in this case the number of negative rois equals to length of extrema array, the number of negative rois is harder to find. Since we look for RoiMap we can't just put 0 as a value there.
CleanShot 2023-08-09 at 11 03 17

src/filters/__tests__/waterShed.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    const roiMapManager = waterShed(image, { threshold: 2 / 255 });
+
+    const resultArray = new Int16Array([
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -2, -2, -2, 0, 0, 0, 0, 0, 0, 0, -2,
+      -2, -2, -2, 0, 0, 0, 0, 0, 0, -2, -2, -2, -2, 0, 0, 0, 0, 0, 0, 0, 0, -2,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -1, 0, 0, 0, 0, 0, 0, 0, 0, -1, -1,
+      -1, 0, 0, 0, 0, 0, 0, 0, -1, -1, -1, -1, 0, 0, 0, 0, 0, -1, -1, -1, -1, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, -1, 0,
+    ]);
```
lpatiny, line 54: We need to find a better way because this is not really readable. I don't know if we have a helper somewhere (I didn't find one).

@targos Is it ok to add a method in 'testUtils' that allows to write

```js
getInt16Array(`
  0, 0, 1, 2
  3, 4, 5, 6,
  7, 8    9,  10
`) and returns a Int16Array ?
```

So something like:
```js
function getInt16Array(string) {
   return new Int16Array(string.split(/[, \r\n]+/.filter(value=>value).map(string=>Number(string)))
}
```

src/filters/__tests__/waterShed.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      map: {
+        data: resultArray,
+        nbPositive: 1,
+        nbNegative: 0,
+        width: 5,
+        height: 5,
+      },
+      whiteRois: [],
+      blackRois: [],
+    });
```
lpatiny, line 94: Please add a test to get the ROIs from the roiMapManager to check if we really have the correct positive and negative one, they have the right position and size.
You need to get the black, white, b/w.

src/filters/__tests__/waterShed.test.ts
targos: We already have helpers to create images, so yeah sure (if it works the same way, or even better if it can reuse the code)

Issue #326, opened 2023-05-10
gaussianBlur: missing validation of sigma options


targos:
If I do this in JS (no TS check):

```
img.gaussianBlur({ size: 25 });
```

I get the error `RangeError: size must be positive and odd` (size is NaN).

PR #382, opened 2023-08-07
refactor: add additional error handling for sigma option and a test case

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/326

### DIFF
--- a/src/filters/gaussianBlur.ts
+++ b/src/filters/gaussianBlur.ts
@@ -20,7 +20,7 @@ export interface GaussianBlurSigmaOptions extends GaussianBlurBaseOptions {
   /**
    * The standard deviation. Specifies the width of the gaussian function in the case where it is the same for x and y.
    */
-  sigma: number;
+  sigma?: number;
   /**
    * Size of the kernel.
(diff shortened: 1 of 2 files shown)


stropitek:
The issue is not about changing the types. The types are correct AFAIK.

The issue is about what happens if the type error is ignored and js code gets executed anyway.

You should add a test that has a `@ts-expect-error`

src/filters/gaussianBlur.ts
```diff
@@ -87,6 +87,10 @@ export function gaussianBlur(
     return separableConvolution(image, kernelX, kernelY, {
       borderType,
     });
+  } else {
+    throw new Error(
+      'sigma is undefined. you must define sigma or sigmaX and sigmaY',
+    );
```
stropitek, line 93: This qualifies as a `TypeError`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

```suggestion
    throw new TypeError(
      'you must either define sigma or sigmaX and sigmaY in the options argument',
    );
```

stropitek (approved):

PR #383, opened 2023-08-08
fix: allow handling floating values

josoriom:


src/draw/drawMarker.ts
targos, line 1: I'm not sure we should round in `drawMarker`. We can let the functions called internally handle float values.

src/draw/drawRectangle.ts
```diff
@@ -56,14 +56,16 @@ export function drawRectangle(
...
     origin = { column: 0, row: 0 },
-    width = image.width,
-    height = image.height,
     strokeColor = getDefaultColor(image),
     fillColor = 'none',
   } = options;
-  const { column, row } = origin;
-
+  const width = Math.round(rectangleWidth ?? image.width);
+  const height = Math.round(rectangleHeight ?? image.width);
```
targos, line 66:
```suggestion
  const width = Math.round(rectangleWidth);
  const height = Math.round(rectangleHeight);
```

You already set the default value above.

src/draw/drawMarker.ts
```diff
@@ -89,8 +89,8 @@ export function drawMarker(
   }
   if (shape === 'square') {
     const origin = {
-      row: Math.floor(point.row - size / 2),
-      column: Math.floor(point.column - size / 2),
+      row: Math.round(point.row - size / 2),
+      column: Math.round(point.column - size / 2),
```
targos, line 93:
```suggestion
      row: point.row - size / 2,
      column: point.column - size / 2,
```

targos (changes requested):

src/draw/drawMarker.ts
```diff
@@ -89,8 +89,8 @@ export function drawMarker(
   }
   if (shape === 'square') {
     const origin = {
-      row: Math.floor(point.row - size / 2),
-      column: Math.floor(point.column - size / 2),
+      row: Math.round(point.row - size / 2),
```
lpatiny, line 92: I would use size-1 for row and col, seems to make more sense to me
`row: Math.round(point.row - (size-1) / 2)`
You want that if the size is 1 the pixel is exactly at the specified position.

src/draw/drawRectangle.ts
```diff
@@ -56,14 +56,16 @@ export function drawRectangle(
...
+    height: rectangleHeight = image.height,
     origin = { column: 0, row: 0 },
-    width = image.width,
-    height = image.height,
     strokeColor = getDefaultColor(image),
     fillColor = 'none',
   } = options;
-  const { column, row } = origin;
-
+  const width = Math.round(rectangleWidth ?? image.width);
```
lpatiny, line 65: rectangleWidth should always be defined and set to the default value image.width so `?? image.width` seems useless.

lpatiny (changes requested):

src/draw/drawRectangle.ts
josoriom: You're right, I did it twice

Issue #320, opened 2023-04-26
operation on images: sum / add


lpatiny:
We also already the method `subtract`.

I could not find the sum of images.

Do we have this method ?

targos:
I don't think we have it. Here's a little article in opencv about it: https://docs.opencv.org/3.4/d0/d86/tutorial_py_image_arithmetics.html

PR #384, opened 2023-08-09
feat: add a sum filter

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/320

### DIFF
--- a/src/compare/add.ts
+++ b/src/compare/add.ts
@@ -8,7 +8,7 @@ import { validateForComparison } from '../utils/validators/validators';
  * @param otherImage - Image to add.
  * @returns The summed image.
  */
-export function sum(image: Image, otherImage: Image): Image {
+export function add(image: Image, otherImage: Image): Image {
   if (image instanceof Image) {
     checkProcessable(image, {
       bitDepth: [1, 8, 16],
--- a/src/compare/add.ts
+++ b/src/compare/add.ts
@@ -12,7 +12,7 @@ import { validateForComparison } from '../utils/validators/validators';
 export function add(image: Image, otherImage: Image): Image {
   if (image instanceof Image) {
     checkProcessable(image, {
-      bitDepth: [1, 8, 16],
+      bitDepth: [8, 16],
       components: [1, 3],
       alpha: false,
     });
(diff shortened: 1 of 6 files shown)


src/compare/sum.ts
```diff
@@ -0,0 +1,33 @@
+import { Image } from '..';
+import checkProcessable from '../utils/validators/checkProcessable';
+import { validateForComparison } from '../utils/validators/validators';
+/**
+ *
+ * Calculate a new image that is the sum between the current image and the otherImage.
+ * @param image - Image to which to add.
+ * @param otherImage - Image to add.
+ * @returns The subtracted image.
```
stropitek, line 9: subtracted?

stropitek (commented):
Can you add tests which show what happens when the sum saturates (sum exceeds max value). Add a test with 8-bit depth and another with 16-bit depth.

I'm not sure if masks should have a `sum` method. It would have the exact same implementation as the `or` method. @lpatiny @targos should we have `sum` be an alias of `or`?

src/compare/sum.ts
```diff
@@ -0,0 +1,33 @@
...
+  validateForComparison(image, otherImage);
+
+  const newImage = image.clone();
+  for (let index = 0; index < image.size; index++) {
+    for (let channel = 0; channel < image.channels; channel++) {
+      const value =
+        image.getValueByIndex(index, channel) +
+        otherImage.getValueByIndex(index, channel);
+
+      newImage.setValueByIndex(index, channel, Math.min(value, image.maxValue));
```
targos, line 29: Should use `clamp` like in all other methods.

targos:
@stropitek not sure why you're mentioning mask here. The PR doesn't touch anything on mask code.

stropitek:
> not sure why you're mentioning mask here. The PR doesn't touch anything on mask code.

It's because the `subtract` method exists on Mask and to be on par we could implement the `sum` method as well. But to me the concept of doing sums and subtractions on masks seems kind of strange. I don't know for which use case the subtract method exists in the first place.

src/compare/add.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Calculate a new image that is the sum between the current image and the otherImage.
+ * @param image - Image to which to add.
+ * @param otherImage - Image to add.
+ * @returns The summed image.
+ */
+export function add(image: Image, otherImage: Image): Image {
+  if (image instanceof Image) {
+    checkProcessable(image, {
+      bitDepth: [1, 8, 16],
```
stropitek, line 15: Support for 1-bit depth can be removed

stropitek:
`image.add(otherImage)` should be possible as well

stropitek (approved):

PR #386, opened 2023-08-16
feat: improve speed of RGBA invert by a factor 3

lpatiny:


lpatiny:
This was just a proof of concept and it does not seem critical (invert is fast anyway).

Issue #406, opened 2023-11-24
check if hypotenuse options are implemented correctly


EscapedGibbon:
```ts
test('testing with rgb image', () => {
  const image = testUtils.createRgbImage([[3, 5, 8]]);
  const otherImage = testUtils.createRgbImage([[4, 12, 15]]);

  const expected = testUtils.createRgbImage([[5, 13, 17]]);
  expect(image.hypotenuse(otherImage, { channels: [0] })).toMatchImage(
    expected,
  );
});
```

when applying hypotenuse only to one channel, I believe this test should not pass, yet it does.

stropitek:
Can you explain what you would expect to happen in this test case?

Are those expectations consistent with what happens in other methods that have the `channels` option?

EscapedGibbon:
> Can you explain what you would expect to happen in this test case?

`channels` option specifies which channels the algorithm should be applied to. So, in this case, it should return `[5,5,8]` and not `[5,13,17].  I checked it now and, somehow, the test works correctly, but if you put `[0,1]` in `channels` options, it will apply hypotenuse on all channels instead of channels 0 and 1.

EscapedGibbon:
> Are those expectations consistent with what happens in other methods that have the `channels` option?

I believe so if I understand what applying the algorithm to specific channels means correctly. There is also `channel` option in `level` function, and I think it works correctly there.

stropitek:
> I believe so if I understand what applying the algorithm to specific channels means correctly. There is also `channel` option in `level` function, and I think it works correctly there.

Can you point me to a test that uses this option for the `level` or another method?

EscapedGibbon:
> > I believe so if I understand what applying the algorithm to specific channels means correctly. There is also `channel` option in `level` function, and I think it works correctly there.
> 
> Can you point me to a test that uses this option for the `level` or another method?

https://github.com/image-js/image-js-typescript/blob/3a9bc791c86cc19f54a738a03221ea9dfac3724f/src/filters/__tests__/level.test.ts#L74

EscapedGibbon:
https://github.com/image-js/image-js-typescript/blob/3a9bc791c86cc19f54a738a03221ea9dfac3724f/src/filters/__tests__/level.test.ts#L95

PR #407, opened 2023-11-29
fix: fix the bug with hypotenuse custom channels

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/406


EscapedGibbon:
I removed `bitDepth` option because it was only used in creating new image. But with `getOutputImage` the bit depth is the same as the input image's.

stropitek (approved):

PR #409, opened 2023-12-01
Fix level default options

opatiny:
- fix: correct level default options
- refactor: rename autoLevel


src/Image.ts
```diff
@@ -942,8 +942,8 @@ export class Image {
    * @param options - Enhance contrast options.
    * @returns The enhanced image.
    */
-  public autoLevel(options: AutoLevelOptions = {}): Image {
-    return autoLevel(this, options);
+  public enhanceContrast(options: IncreaseContrastOptions = {}): Image {
```
targos, line 945:
```suggestion
  public increaseContrast(options: IncreaseContrastOptions = {}): Image {
```

Issue #408, opened 2023-12-01
typedoc issue


EscapedGibbon:
There is a bunch of functions that is not documented by typedoc, even though they are exported.
Notably,
- getFeret
- getMbr
- getConvexHull
- getExtrema
- filterPoints
etc.

PR #411, opened 2023-12-02
chore: make all types that are used or returned by the public interface also public

stropitek:
Closes: https://github.com/image-js/image-js-typescript/issues/408

Based on warnings produced by typedoc.


stropitek:
I will be also adding an option to typedoc-action so that it can treat warnings as errors and that those warnings can be caught early in our PRs.

stropitek:
https://github.com/zakodium/typedoc-action/pull/22

My idea is to do this:

- In PR: treat warnings as errors and skip the step that publishes the documentation
- On the main branch: do not treat warnings as errors and publish the documentation

PR #412, opened 2023-12-04
chore: add additional exports for elements to be seen by typedoc

EscapedGibbon:
add exports for filterPoints and getExtrema so they could be added to typedoc generated documentation.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ export * from './draw';
 export * from './featureMatching';
 export * from './filters';
 export * from './geometry';
+export * from './utils/geometry/filterPoints';
 export * from './Image';
 export * from './load';
 export * from './Mask';
--- a/src/roi/waterShed.ts
+++ b/src/roi/waterShed.ts
@@ -3,7 +3,7 @@ import PriorityQueue from 'js-priority-queue';
 import { RoiMapManager } from '..';
 import { Image } from '../Image';
 import { Mask } from '../Mask';
-import getExtrema from '../compute/getExtrema';
+import { getExtrema } from '../compute/getExtrema';
 import { Point } from '../geometry';
 import { filterPoints } from '../utils/geometry/filterPoints';
 import checkProcessable from '../utils/validators/checkProcessable';
--- a/src/compute/__tests__/getExtrema.test.ts
+++ b/src/compute/__tests__/getExtrema.test.ts
@@ -1,4 +1,4 @@
-import getExtrema from '../getExtrema';
+import { getExtrema } from '../getExtrema';
 
 test('minimum of grey image from legacy code', () => {
   const image = testUtils.createGreyImage([
--- a/src/compute/getExtrema.ts
+++ b/src/compute/getExtrema.ts
@@ -2,7 +2,7 @@ import { Image, Mask, Point } from '..';
 import { assertUnreachable } from '../utils/validators/assert';
 import checkProcessable from '../utils/validators/checkProcessable';
 
-interface ExtremaOptions {
+export interface ExtremaOptions {
   /**
    * Chooses what kind of extremum to compute.
    * @default `'maximum'`
@@ -30,10 +30,7 @@ interface ExtremaOptions {
  * @param options - ExtremaOptions
  * @returns Array of Points.
  */
-export default function getExtrema(
-  image: Image,
-  options: ExtremaOptions,
-): Point[] {
+export function getExtrema(image: Image, options: ExtremaOptions): Point[] {
   const { kind = 'maximum', mask, algorithm = 'star', maxEquals = 2 } = options;
   checkProcessable(image, {
     bitDepth: [8, 16],
--- a/src/compute/index.ts
+++ b/src/compute/index.ts
@@ -1,3 +1,4 @@
 export * from './mean';
 export * from './histogram';
 export * from './median';
+export * from './getExtrema';
--- a/src/utils/geometry/__tests__/filterPoints.test.ts
+++ b/src/utils/geometry/__tests__/filterPoints.test.ts
@@ -1,4 +1,4 @@
-import getExtrema from '../../../compute/getExtrema';
+import { getExtrema } from '../../../compute/getExtrema';
 import { filterPoints } from '../filterPoints';
 
 test('combine minimum points after getExtrema function', () => {
--- a/src/utils/geometry/filterPoints.ts
+++ b/src/utils/geometry/filterPoints.ts
@@ -2,7 +2,7 @@ import { Image } from '../..';
 
 import { Point } from './points';
 
-interface FilterPointsOptions {
+export interface FilterPointsOptions {
   /**
    * The number of points that should be removed if they are close to extremum.
    * @default `0`


src/index.ts
```diff
@@ -6,6 +6,7 @@ export * from './draw';
 export * from './featureMatching';
 export * from './filters';
 export * from './geometry';
+export * from './utils/geometry/filterPoints';
```
targos, line 9: We should not expose lower-level files from here. If `filterPoints` is meant to be public API (but should it really? The name is very generic for a much more specific function), it should be exported from the `geometry/index.ts` file.

stropitek: I also have an issue with the name and wasn't sure if it made sense to export it from the library vs let the algo be run with certain options of the watershed function.

If this is meant to be used to get starting points for watershed algorithm, why not either name it after that (like `findWatershedPoints` or find something in the middle in terms of genericness (like `getFilteredLocalExtrema`).


@lpatiny WDYT? I believe you discussed this with @EscapedGibbon

stropitek:
> add exports for filterPoints and getExtrema so they could be added to typedoc generated documentation.

The main reason for this change is to make those functions available publicly for anyone who installed the library to use. It being in typedoc is a side effect of that.

lpatiny:
Maybe we could rename the method filterPoints to be lest generic and actually give as name what is does: `removeClosePoints`.

```
  const points = getExtrema(image, {kind: 'maximum'});
  const distinctPoints = removeClosePoints(points, {distance: 5, kind: 'maximum'});
```

src/roi/__tests__/waterShed.test.ts
```diff
@@ -70,8 +69,8 @@ describe('Test WaterShed Roi generation', () => {
       blackRois: [],
     });
     const rois = roiMapManager.getRois({ kind: 'bw' });
-    expect(rois[0].origin).toEqual({ column: 5, row: 5 });
-    expect(rois[1].origin).toEqual({ column: 2, row: 1 });
```
stropitek, line 74: There are too many things that are not tested here. The number of rois and the dimensions of the roi for example.

src/roi/__tests__/waterShed.test.ts
```diff
@@ -168,16 +167,16 @@ describe('Test WaterShed Roi generation', () => {
     const mask = image.threshold({ algorithm: 'otsu' });
```
stropitek, line 167: We should start from a mask and not create it from an image, because we are not testing otsu in this test, only watershed

src/utils/geometry/__tests__/removeClosePoints.test.ts
```diff
@@ -1,32 +1,47 @@
...
+    [1, 1, 1],
+    [1, 6, 6],
+    [1, 1, 1],
   ]);
+  const points = [
+    { column: 2, row: 1 },
+    { column: 2, row: 2 },
+  ];
+
+  const result = removeClosePoints(points, image, { distance: 2 });
```
stropitek, line 42: Would this have the same result with a value of 1?

src/roi/__tests__/waterShed.test.ts
stropitek: Actually, this should not be tested. Some reason as below, we already checked the roiMapManager is correct.

src/roi/__tests__/waterShed.test.ts
```diff
@@ -192,6 +191,6 @@ describe('Test WaterShed Roi generation', () => {
     });
     const rois = roiMapManager.getRois({ kind: 'bw' });
     expect(rois[0].origin).toEqual({ column: 0, row: 0 });
-    expect(rois[0].surface).toEqual(60);
```
stropitek, line 195: ditto, checking result of getRois is out of scope for this test

src/utils/geometry/__tests__/removeClosePoints.test.ts
EscapedGibbon: No, because the function checks if the euclidean distance between points is strictly smaller than the parameter value. Should it be smaller or equal?

stropitek: No I think it's better the way it is now

src/roi/__tests__/computeRois.test.ts
```diff
@@ -15,3 +16,23 @@ test('3x3 mask', () => {
...
+    [4, 3, 3, 3, 3, 3, 2, 2, 2, 3],
+    [4, 4, 3, 3, 3, 3, 2, 2, 2, 2],
+    [4, 4, 4, 4, 3, 2, 2, 2, 2, 3],
+    [4, 4, 4, 4, 3, 3, 3, 3, 2, 3],
+  ]);
+
+  const roiMapManager = waterShed(image, { threshold: 2 / 255 });
+  const rois = roiMapManager.getRois({ kind: 'bw' });
+  expect(rois[1].origin).toEqual({ column: 5, row: 5 });
+  expect(rois[0].origin).toEqual({ column: 2, row: 1 });
```
stropitek, line 37: Can you also test what the map manager returns instead of what rois are generated from that?

src/utils/geometry/removeClosePoints.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Finds extreme values of an image which are not stacked together.
+ * @param points - Array of points that should be combined to improve.
+ * @param image - Image which extrema are calculated from.
+ * @param options - FilterPointsOptions
+ * @returns Array of Points.
+ */
+export function removeClosePoints(
+  points: Point[],
+  image: Image,
+  options?: RemoveClosePointsOptions,
```
stropitek, line 32: Let's make this mandatory, and also make the `distance` and `kind` options mandatory.

It doesn't make sense to have defaults that do nothing.

src/utils/geometry/__tests__/removeClosePoints.test.ts
```diff
@@ -1,32 +1,47 @@
...
+    { column: 1, row: 1 },
+    { column: 1, row: 4 },
+    { column: 3, row: 1 },
+    { column: 2, row: 4 },
+  ];
 
-  const result = filterPoints(points, image, {
-    removeClosePoints: 2,
+  const result = removeClosePoints(points, image, {
+    distance: 2.5,
```
stropitek, line 21: Same result with 3? Let's put 3 here then

stropitek (approved):
LGTM after final comments are resolved

src/utils/geometry/removeClosePoints.ts
EscapedGibbon: Would it make sense to make the `kind` option mandatory for `getExtrema` as well?

stropitek: I don't have a strong opinion. It certainly makes more sense to have a default for this option in `getExtrema` since it's kind of implied that this option exists.

In `removeClosePoints` it's really not obvious that such an options should exist

EscapedGibbon: Ok, then I believe the changes are done.

src/roi/waterShed.ts
```diff
@@ -77,7 +77,7 @@ export function waterShed(
       kind: 'minimum',
       mask,
     });
-    points = removeClosePoints(points, image, { kind: 'minimum' });
+    points = removeClosePoints(points, image, { kind: 'minimum', distance: 0 });
```
stropitek, line 80: If calling this with a distance of 0 does not filter any points, why call it?

src/roi/__tests__/computeRois.test.ts
```diff
@@ -32,7 +32,33 @@ test('test 2, waterShed for a grey image', () => {
...
+           0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+          -1, -1, -1,  0,  0,  0,  0,  0,  0,  0, -1, -1,
+          -1, -1,  0,  0,  0,  0,  0,  0, -1, -1, -1, -1,
+           0,  0,  0,  0,  0,  0,  0,  0, -1,  0,  0,  0,
+           0,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0,  0,
+           0,  0,  0,  0,  0,  0, -2, -2, -2,  0,  0,  0,
+           0,  0,  0,  0, -2, -2, -2, -2,  0,  0,  0,  0,
+           0, -2, -2, -2, -2,  0,  0,  0,  0,  0,  0,  0,
+           0,  0, -2,  0
+        `),
```
stropitek, line 47: It is really hard to try to understand this test when it written out with a number of columns which is different from the number of columns of the image the result is based on.

EscapedGibbon: CleanShot 2023-12-14 at 10 47 07

I have a question regarding this test.  I check ROIs for their origins but the second ROI starts before actual ROI points(column: 5 row: 5). It is due to the fact that computeRois checks minimum column and minimum row separately. In this case the origin forms a point that doesn't technicall belong to ROI. My question is this an acceptable behavior?

CleanShot 2023-12-14 at 10 49 44

stropitek: That is normal. I asked Luc and he also thinks it is normal and expected.

The origin does not have to be a point of the ROI. Maybe it's just a terminology issue, but I cannot think of a better term for this.

stropitek (approved):

PR #416, opened 2023-12-15
Start implementing stacks

opatiny:
- feat: basic Stack API
- feat(Stack): implement constructor
- feat(Stack): implement functions to get values
- feat(Stack): implement minImage
- feat(Stack): implement maxImage
- feat(Stack): implement iterator
- test(Stack): add min and max image tests
- feat(Stack): implement getStackFromFolder for debug
- fix(Stack): enhance Stack properties
- feat(Stack): implement map and filter
- feat(Stack): implement meanImage
- feat(Stack): implement medianImage
- test: enhance stack coverage


src/stack/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './maxImage';
+export * from './minImage';
```
targos, line 2: what about median and mean?

src/stack/meanImage.ts
```diff
@@ -0,0 +1,54 @@
...
+      const index = i * stack.channels + channel;
+      meanArray[index] = sum[index] / stack.size;
+    }
+  }
+
+  return new Image(image.width, image.height, {
+    data: meanArray,
+    colorModel: stack.colorModel,
+    bitDepth: image.bitDepth,
+  });
```
targos, line 53: Do you need all that? I would expect a `Image.createFrom` like the min/max implementations.

PR #433, opened 2024-02-21
feat: make encode types easier to use

stropitek:


src/save/write.ts
```diff
@@ -123,25 +118,16 @@ function getDataToWrite(
...
   if (!options || !('format' in options)) {
     const extension = nodePath.extname(destinationPath).slice(1).toLowerCase();
-    if (extension === 'png') {
-      format = 'png';
-      return encode(image, { format });
-    } else if (extension === 'jpg' || extension === 'jpeg') {
-      format = 'jpg';
-      return encode(image, { format });
+    if (extension === 'png' || extension === 'jpg' || extension === 'jpeg') {
+      return encode(image, { format: extension });
```
targos, line 124: I realize that options should be spread here, to pass the specific encoder options.

targos (approved):

Issue #436, opened 2024-03-15
Add `cropRectangle` function


targos:
Takes an array of 4 points, validates, and extracts the closest possible rectangle from it (Math.ceil for width, height), rotating around the axis with the smallest angle.

PR #439, opened 2024-03-20
add cropRectangle

stropitek:
Closes: https://github.com/image-js/image-js-typescript/issues/436


src/Image.ts
```diff
@@ -109,6 +109,10 @@ import {
   threshold,
   ThresholdOptions,
 } from './index';
+import {
+  cropRectangle,
+  CropRectangleOptions,
+} from './operations/cropRectangle';
```
targos: Should be imported from `./operations`

src/Image.ts
```diff
@@ -807,6 +811,13 @@ export class Image {
     return crop(this, options);
   }
 
+  /**
+   *
+   */
```
targos: Missing docs

stropitek:
The correctness of the expected test results was checked by plotting the points from which we are supposed to interpolate and check visually if the value corresponds to the nearest neighbour.

targos (approved):

Issue #438, opened 2024-03-20
Add transform tests


stropitek:
Currently only a translation is tested against opencv.

We should add some tests with rotation and scaling as well, see if opencv's implementation matches ours.

MDSALMANSHAMS:
Hi @stropitek, I have been doing the transformation based image processing in JS for some months (in my projects), and I would love to contribute in here. (just need to go through TS :)).

stropitek:
Hello @MDSALMANSHAMS 

The comment you made is out of context and unactionnable. Please refrain from doing such comments in the future. I recommend that you read https://opensource.guide/how-to-contribute/ to understand how open source contributions work.

PR #441, opened 2024-04-03
test: add testing cases for transform function

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/438


stropitek:
Please document the opencv call that was used to generate the opencv images

stropitek:
Why is this still in draft?

You can merge with main to get rid of the CI errors

targos:
@EscapedGibbon What commands/actions did you do **exactly** to end up with these three new commits in your branch?

EscapedGibbon:
> @EscapedGibbon What commands/actions did you do **exactly** to end up with these three new commits in your branch?

git fetch+git rebase origin/main

EscapedGibbon:
> Ok, now I suggest you use `git merge origin/main` and `git push` to avoid any issues. `git rebase` is a nice tool but shouldn't be used if it can be avoided.

Ok, got it.

stropitek (commented):
Please also test without `fullImage=true` so that the matrix is closer to what you did with opencv.

EscapedGibbon:
The transform function doesn't work exactly the way it is supposed to, I think.
Here is the transformation:
```js
let imageReflect = image.transform(
  [
    [1, 0, 0],
    [0, -1, 0],
  ],
  { inverse: false, fullImage: true },
);
```

CleanShot 2024-04-17 at 13 49 27

EscapedGibbon:
Same thing with scaling:
CleanShot 2024-04-17 at 13 55 47

stropitek:
Our test result is in line with my mental model of how transform works, and to make it work we should use `height-1` in the matrix instead of `height`. But there is maybe something I am missing since opencv does not create the same result.

Can you use `warpAffine` instead of `warpPerspective` like in the translation test?

targos:
As a reference, here are the Python scripts that were used to generate the test images: https://github.com/image-js/image-js/tree/master/test/python

EscapedGibbon:
As was discussed, I looked at the border types and opencv parameters and they are normally the same as in image-js(borderType = "constant", borderValue=0,interpolationType=bilinear). The scaling test is passing now, but it passes only if I scale by 2 or by 4. In other cases the output differs from the opencv image. The one with affine transformation doesn't work as well, and it's not the issue of border or interpolation type.

stropitek:
CleanShot 2024-04-22 at 14 48 51


Does the reporter report just the first mismatch in the image or all the wrong pixels?

In other words is the reported wrong pixel the only one that is wrong?

EscapedGibbon:
> CleanShot 2024-04-22 at 14 48 51
> 
> Does the reporter report just the first mismatch in the image or all the wrong pixels?
> 
> In other words is the reported wrong pixel the only one that is wrong?

No, it's not the only one. It's just the first one.

targos:
@EscapedGibbon We looked into OpenCV's implementation of the bilinear interpolation yesterday (https://github.com/opencv/opencv/blob/4.x/modules/imgproc/src/opencl/warp_affine.cl#L123-L129) and it is actually not exactly correct, for performance reasons. So it's normal that we don't get the exact same results and I added a margin of error in https://github.com/image-js/image-js-typescript/pull/441/commits/b8b61ca40fc22f21cabc15bfa3b4eae34147a4aa

targos (approved):

PR #445, opened 2024-04-10
refactor: rename bitDepth to depth

### LINKED ISSUE #444: Exporting 16-bits images as png doesn't work
In the function `encodePng` that uses `fast-png` library : 
```
encode(image.getRawImage(), options);
```
We have :
```
Image.getRawImage(): {
    width: number;
    height: number;
    data: ImageDataArray;
    channels: number;
    bitDepth: BitDepth;
}
```

While : 
```
export interface ImageData {
    width: number;
    height: number;
    data: PngDataArray;
    depth?: BitDepth;
    channels?: number;
}
```

`bitDepth` in `getRawImage` has to be changed to `depth`
So here since we don't pass a value to `depth` it uses the default value in `fast-png` which is 8 bits

moonayyur:
closes #444


stropitek:
Would it be simpler to remap `bitDepth` to `depth` when passing the options to fast-png?

targos (changes requested):
I'm sorry but I don't see a good reason for making this big change.

stropitek:
> Would it be simpler to remap `bitDepth` to `depth` when passing the options to fast-png?

Simpler but mainly non-breaking

src/save/encodePng.ts
```diff
@@ -14,5 +14,14 @@ export function encodePng(
...
 ): Uint8Array {
-  return encode(image.getRawImage(), options);
+  return encode(
+    {
+      data: image.getRawImage().data,
+      width: image.getRawImage().width,
+      height: image.getRawImage().height,
+      channels: image.getRawImage().channels,
+      depth: image.getRawImage().bitDepth,
+    },
```
targos, line 24: Can you avoid the repetition with a variable?

stropitek:
You can merge with main to get rid of the CI errors

targos:
Can you please add a test that would fail before the fix?

targos (changes requested):
You fix the `encodePng` function, so the test should execute that function, not the one from `fast-png` directly.

moonayyur:
> You fix the `encodePng` function, so the test should execute that function, not the one from `fast-png` directly.

So the image must be converted to 8 bits first ? Since the `encodePng` function only accepts a prop of type `Image` where `bitDepth` is mandatory (so we can't test directly with the default value of `fast-png`)

targos:
I don't see why it would have to be converted to 8 bits first. `bitDepth` is mandatory, but its value doesn't have to be 8.

moonayyur:
Oh, okay I had misunderstood this 
> Can you please add a test that would fail before the fix?

I thought I should write a test that shows the previous error

targos:
You should write a test that pass with the fix and doesn't pass without the fix.

moonayyur:
The previous error was that the attribute `depth` in `fast-png` wasn't getting any value from `img` since it was named `bitDepth`
But now `bitDepth` is renamed to `depth` inside `encodePng`, so I don't see how I could go back to the error in the test

targos:
You're correct, and that's not what I'm asking.
Somehow I'm not making myself clear but I don't know how to say it differently 😕

targos (approved):

PR #448, opened 2024-04-11
feat: add option to specify the number of slots for the histogram

### LINKED ISSUE #447: Threshold on 16 bits images is too slow
The calculations of threshold based on the histogram is too slow if the images are 16 bits.

https://github.com/image-js/image-js-typescript/blob/main/src/operations/threshold.ts#L84

We should at least have an options to specify the number of slots to use for the histogram calculation and for me the default value should be 2**8.

The result by itself should barely change.

moonayyur:
closes #447

### DIFF
--- a/src/operations/threshold.ts
+++ b/src/operations/threshold.ts
@@ -73,15 +73,14 @@ export type ThresholdOptions =
 /**
  * Compute threshold value for an image using the specified algorithm.
  * @param image - The grey image.
- * @param algorithm - Algorithm that defines the threshold.
- * @param slots - The number of slots that histogram can have.
+ * @param options - Threshold options.
  * @returns The threshold value for the image.
  */
 export function computeThreshold(
   image: Image,
-  algorithm: ThresholdAlgorithm = 'otsu',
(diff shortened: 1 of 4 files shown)


src/operations/threshold.ts
```diff
@@ -70,18 +74,20 @@ export type ThresholdOptions =
  * Compute threshold value for an image using the specified algorithm.
  * @param image - The grey image.
  * @param algorithm - Algorithm that defines the threshold.
+ * @param slots - The number of slots that histogram can have.
  * @returns The threshold value for the image.
  */
 export function computeThreshold(
   image: Image,
   algorithm: ThresholdAlgorithm = 'otsu',
+  slots?: number,
```
lpatiny, line 83: @stropitek @targos : We have now 2 optional arguments ... I would make a breaking change to have the function:
```js
export function computeThreshold(image, options = {})
const {algorithm = 'otsu', slots} = options;
```

targos: sgtm

lpatiny (approved):

PR #450, opened 2024-04-18
fix: scale threshold when slots number is changed

moonayyur:
#447

### DIFF
--- a/src/operations/threshold.ts
+++ b/src/operations/threshold.ts
@@ -87,7 +87,7 @@ export function computeThreshold(
     );
   }
   const histogram = image.histogram({ slots });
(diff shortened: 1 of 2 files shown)


src/operations/threshold.ts
```diff
@@ -87,38 +87,39 @@ export function computeThreshold(
     );
   }
   const histogram = image.histogram({ slots });
+  const scale = options.slots ? 2 ** image.bitDepth / options.slots : 1;
```
lpatiny, line 90: @targos We didn't have a rule that forbid us to use options.slots when it has been destructured ? Or it is because we send it to image.histogram and eslint does not now it can not be changed by inside the function ?

@moonayyur I think you can use slots and not options.slots.

lpatiny (approved):

PR #453, opened 2024-04-26
wip: fix: resize with interpolationType nearest

tpoisseau:
Closes: https://github.com/image-js/image-js-typescript/issues/452

### DIFF
--- a/src/geometry/__tests__/resize.test.ts
+++ b/src/geometry/__tests__/resize.test.ts
@@ -1,13 +1,23 @@
-import { encodePng } from '../../save';
+import * as path from 'node:path';
 
-test.skip('compare result of resize with opencv (nearest)', () => {
+import { encodePng, write } from '../../save';
+
+test('compare result of resize with opencv (nearest)', async () => {
   const img = testUtils.load('opencv/test.png');
+  const expectedImg = testUtils.load('opencv/testResizeNearest.png');
 
   const resized = img.resize({
     xFactor: 10,
     yFactor: 10,
     interpolationType: 'nearest',
   });
+
+  const substraction = expectedImg.clone().subtract(resized);
+  await write(
+    path.join(__dirname, 'resize_nearest_substraction.png'),
+    substraction,
+  );
+  await write(path.join(__dirname, 'resize_nearest.png'), resized);
 
   expect(resized).toMatchImage('opencv/testResizeNearest.png');
 });
--- a/src/geometry/resize.ts
+++ b/src/geometry/resize.ts
@@ -58,17 +58,21 @@ export function resize(image: Image, options: ResizeOptions): Image {
     borderType = 'constant',
     borderValue = 0,
   } = options;
-  const { width, height } = checkOptions(image, options);
+  const { width, height, xFactor, yFactor } = checkOptions(image, options);
   const newImage = Image.createFrom(image, { width, height });
   const interpolate = getInterpolationFunction(interpolationType);
   const interpolateBorder = getBorderInterpolation(borderType, borderValue);
   const clamp = getClamp(newImage);
-  const intervalX = (image.width - 1) / (width - 1);
-  const intervalY = (image.height - 1) / (height - 1);
+  const wRatio = 1 / xFactor;
+  const hRatio = 1 / yFactor;
+  // const intervalX = (image.width - 1) / (width - 1);
+  // const intervalY = (image.height - 1) / (height - 1);
   for (let row = 0; row < newImage.height; row++) {
     for (let column = 0; column < newImage.width; column++) {
-      const nx = column * intervalX;
-      const ny = row * intervalY;
+      const nx = (column + 0.5) * wRatio;
+      const ny = (row + 0.5) * hRatio;
+      // const nx = column / xFactor;
+      // const ny = row / yFactor;
       for (let channel = 0; channel < newImage.channels; channel++) {
         const newValue = interpolate(
           image,
--- a/src/utils/interpolatePixel.ts
+++ b/src/utils/interpolatePixel.ts
@@ -61,8 +61,8 @@ function interpolateNearest(
   channel: number,
   interpolateBorder: BorderInterpolationFunction,
 ): number {
-  column = Math.round(column);
-  row = Math.round(row);
+  column = Math.floor(column);
+  row = Math.floor(row);
 
   return interpolateBorder(column, row, channel, image);
 }


src/utils/interpolatePixel.ts
```diff
@@ -61,8 +61,8 @@ function interpolateNearest(
   channel: number,
   interpolateBorder: BorderInterpolationFunction,
 ): number {
-  column = Math.round(column);
-  row = Math.round(row);
+  column = Math.floor(column);
+  row = Math.floor(row);
```
targos, line 65: Intuitively I'd say that the solution shouldn't involve changing the interpolation functions, because they are validated by other algorithms like rotate and transform.
Resize has the same issue with all interpolations which makes me think the problem is located only in the resize.ts file.

targos:
@stropitek wdyt? Maybe we should remove the custom code and use `transform` with a scaling matrix?

src/utils/interpolatePixel.ts
tpoisseau: I agree, I am currently reworking to this part.
I rename actuals `interpolateX` functions to `interpolateNeighbourX`. And provide `interpolatePositionX` functions.

So this algorithm change will be applied only on `nearest` interpolation. And will avoid to recompute `ny` on each `row` iteration

stropitek:
I think it would make sense to use transform to have a consistent implementation of methods like `resize`, `cropRectangle`, `rotate` which all just iterate over a target image, calculate a position on the source image and then interpolate the value.

tpoisseau:
I would like to try to use transform matrix instead but I don't know how it work

targos:
https://en.wikipedia.org/wiki/Scaling_(geometry)

With a 2x3 matrix, it should be 

```
[
  [ factor   0    0 ]
  [   0    factor 0 ]
]

**Comment by tpoisseau** [comment]
It's the best I can do.
Nearest is working.
Bilinear is very close from the opencv result.

I let you review before resetting snapshots to let the tests pass.

**Comment by tpoisseau** [comment]
Bilinear PR result:
resize_bilinear
resize_bilinear_substraction

Bilinear main result:
resize_bilinear
resize_bilinear_substraction

Bilinear OpenCV:
testResizeBilinear

Nearest PR result:
resize_nearest
resize_nearest_substraction

Nearest main result:
resize_nearest
resize_nearest_substraction

Nearest OpenCV:
testResizeNearest

**Comment by targos** [comment]
Bilinear is known to give different result compared to OpenCV. Solution is to add a margin of error with comment like here:
https://github.com/image-js/image-js-typescript/blob/8d81c2b4094c25522b9c40aec4b6469837123352/src/geometry/__tests__/transform.test.ts#L110-L113

Use the smallest possible error margin

**Comment by tpoisseau** [comment]
> Do we still need custom implementation of the interpolation functions?

I removed It because I do not use it anymore.

> Bilinear is known to give different result compared to OpenCV.

Ok I thought we should be one to one with OpenCV (And I found his less precise algorithm more smoother).

---

If you agree with theses code changes I think we can reset snapshots and merge

**Comment by targos** [comment]
> I found his less precise algorithm more smoother

Really? The difference is barely visible in the other tests (you have to quickly change from one image to the other to see that the colors change, but they are basically the same).

File: src/utils/interpolatePixel.ts [hunk 2]
```diff
@@ -46,97 +64,93 @@ export function getInterpolationFunction(
...
 }
 
 /**
  * Interpolate using bicubic interpolation.
  * @param image - The image to interpolate.
- * @param column - Column index.
- * @param row - Row index.
- * @param channel - Channel index.
+ * @param nx - interpolation of original column (float) from target column.
+ * @param ny - interpolation of original row (float) from target row.
```
stropitek, line 131: ditto

src/utils/interpolatePixel.ts
```diff
@@ -12,32 +12,50 @@ export const InterpolationType = {
 export type InterpolationType =
   (typeof InterpolationType)[keyof typeof InterpolationType];
 
-type InterpolationFunction = (
+type InterpolationNeighbourFunction = (
```
targos, line 15: We should use "neighbour" or "neighbor", not a mix of both (consistency).

tpoisseau:
> The difference is barely visible in the other tests
I did not check other tests, but take a lot of time to tweak trying to get result most similar result with opencv.
Comparing result with substraction, zoom-in and pixel grid ^^'

tpoisseau:
After snaptshots reset, some tests did not pass yet

targos:
There are still substantial changes to interpolateNeighbourBilinear which may explain it.

tpoisseau:
Yes, I'll adapt tests data

targos:
But are you sure about the change to the algorithm? It worked fine until now.

tpoisseau:
I'm not sure.
I found bilinear for the small image better and more similar to opencv. but for largeur image, black and white (like letters with rotation and matching) it seems more blurry. rotation seems to produce some color shifting with triangles.

Maybe I should rollback the bilinear but keep nearest.

src/geometry/resize.ts
```diff
@@ -53,36 +49,26 @@ export interface ResizeOptions {
  * @returns The new image.
  */
 export function resize(image: Image, options: ResizeOptions): Image {
+  const { interpolationType = 'bilinear' } = options;
   const {
-    interpolationType = 'bilinear',
-    borderType = 'constant',
+    borderType = interpolationType === 'bilinear' ? 'replicate' : 'constant',
```
targos, line 54: Why? Is it OpenCV behavior?

tpoisseau: I don't know but bilinear was better with replicate by default.

targos: This is probably not the right way to fix it.

src/geometry/__tests__/resize.test.ts
```diff
@@ -1,26 +1,46 @@
...
-  expect(resized).toMatchImage('opencv/testResizeBilinear.png');
+  const substraction = expectedImg.clone().subtract(resized);
+  await write(
+    path.join(__dirname, 'resize_bilinear_substraction.png'),
+    substraction,
+  );
+  await write(path.join(__dirname, 'resize_bilinear.png'), resized);
+
+  // OpenCV bilinear interpolation is less precise for speed.
+  expect(resized).toMatchImage('opencv/testResizeBilinear.png', { error: 25 });
```
targos, line 43: `error: 25` is too much (that's 10% of the range of values 0-255)

PR #454, opened 2024-05-02
fix: resize nearest

tpoisseau:
Refs: https://github.com/image-js/image-js-typescript/issues/452

---

chirurgical fix, rework of https://github.com/image-js/image-js-typescript/pull/453

previous PR had lot of issues due to fix for bilinear interpolation breaking lot of tests for general worse results (only clear better results was for the opencv bilinear test).

better / worse are personal appreciation.

This PR fix resize with nearest interpolation by using transform matrix without change `interpolateNearest` function and adding small translation to compensate.
Changing interpolation function would break some tests but not for the best. I continue to think more work need to be done on this topic.

- interpolateNearest should use floor I guess
- compensate with translation may not be the best things to do
- other resize interpolation should also use transform
- maybe the algorithm for transform is not exactly right or be specific for each interpolation method

It's a lot of work to plan for really small, nearly impossible to see with the naked eye and lot of decision and estimation to do to actualize breaking tests or not.

### DIFF
--- a/src/geometry/__tests__/resize.test.ts
+++ b/src/geometry/__tests__/resize.test.ts
@@ -4,20 +4,12 @@ import { encodePng, write } from '../../save';
 
 test('compare result of resize with opencv (nearest)', async () => {
   const img = testUtils.load('opencv/test.png');
-  const expectedImg = testUtils.load('opencv/testResizeNearest.png');
 
   const resized = img.resize({
     xFactor: 10,
     yFactor: 10,
     interpolationType: 'nearest',
   });
-
-  const substraction = expectedImg.clone().subtract(resized);
-  await write(
-    path.join(__dirname, 'resize_nearest_substraction.png'),
-    substraction,
-  );
-  await write(path.join(__dirname, 'resize_nearest.png'), resized);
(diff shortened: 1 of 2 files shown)


targos (changes requested):
It can ref https://github.com/image-js/image-js-typescript/issues/452, but it doesn't close it (only nearest-neighbor is fixed).

tpoisseau:
Ok for refs instead closes.

src/geometry/__tests__/resize.test.ts
```diff
@@ -1,25 +1,43 @@
...
   const img = testUtils.load('opencv/test.png');
+  const expectedImg = testUtils.load('opencv/testResizeNearest.png');
 
   const resized = img.resize({
     xFactor: 10,
     yFactor: 10,
     interpolationType: 'nearest',
   });
 
+  const substraction = expectedImg.clone().subtract(resized);
```
targos, line 15: Sorry it seems I forgot to submit the comment on my previous review: please remove the debugging code.

targos (approved):

PR #459, opened 2024-05-08
feat: add multiply/divide functions

EscapedGibbon:


EscapedGibbon:
I haven't added functions as image methods. Also right now functions do not multiply alpha channels, maybe it is not needed at all, or maybe we can add an option if alpha channel should be taken into account as well.

src/compute/multiply.ts
```diff
@@ -0,0 +1,49 @@
...
+  image: Image,
+  value: number,
+  options: MultiplyOptions = {},
+) {
+  const {
+    channels = new Array(image.components).fill(0).map((value, index) => index),
+  } = options;
+  validateChannels(channels, image);
+
+  const newImage = getOutputImage(image, options, { clone: true });
```
stropitek, line 35: Why didn't you provide the `out` option which let's you choose where to put the result, in the same way we do in many other processing methods?

src/compute/divide.ts
```diff
@@ -0,0 +1,42 @@
...
+  image: Image,
+  value: number,
+  options: MultiplyOptions = {},
+) {
+  const {
+    channels = new Array(image.components).fill(0).map((value, index) => index),
+  } = options;
+  validateChannels(channels, image);
+  if (value === 0) {
+    throw new Error(`Value cannot be equal to 0.`);
```
stropitek, line 26:
```suggestion
    throw new TypeError(`Cannot divide by 0`);
```

1. Use TypeError when you receive a bad argument
2. We don't put periods in our error messages.
3. Try to write the message in a way that the error is easy to interpret by the user.

src/compute/__tests__/divide.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  ]);
+  image = divide(image, 2);
+  const result = testUtils.createRgbaImage([
+    [115, 40, 60, 255],
+    [50, 70, 6, 1],
+  ]);
+  expect(image).toStrictEqual(result);
+});
+
+test('must throw error', () => {
```
stropitek, line 16:
```suggestion
test('error when dividing by 0', () => {
```

The name of the test is too generic

src/compute/multiply.ts
EscapedGibbon: I looked at several functions as examples(level, invert, pixelate) so I did it the same way. What did I do wrong?

src/compute/divide.ts
```diff
@@ -0,0 +1,42 @@
...
+
+//The used options are the same as in multiply function.
+import { MultiplyOptions } from './multiply';
+
+/**
+ *
+ * Divides image pixels by a certain value.
+ * @param image - image to which division will be applied.
+ * @param value - Value by which each pixel will be divided.
+ * @param options - Multiply options
```
stropitek, line 13: Divide not multiply

src/compute/multiply.ts
stropitek: Sorry I missed something while reviewing

src/compute/multiply.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (channels.length === 0) {
+    return newImage;
+  }
+  for (let channel = 0; channel < channels.length; channel++) {
```
stropitek, line 39: It's a bit confusing to name this `channel` when it is actually the index in which to look for the channel.

src/compute/multiply.ts
```diff
@@ -0,0 +1,49 @@
...
+    for (let row = 0; row < newImage.height; row++) {
+      for (let column = 0; column < newImage.width; column++) {
+        const newIntensity =
+          newImage.getValue(column, row, channels[channel]) * value;
+        newImage.setClampedValue(column, row, channel, newIntensity);
```
stropitek, line 44: isn't the third parameter wrong here?

stropitek:
> Also right now functions do not multiply alpha channels

Is it not possible by providing the alpha channel in the `channels` option?

stropitek:
We need at least one test to test each option (`channels` and `out`)

stropitek:
You can also add `divide` and `multiply` as `Image` methods

src/compute/divide.ts
EscapedGibbon: It should be, but the interface I use is from MultiplyOptions. In the parameters docs should I put that it is divide options anyway?

targos: You should not use MultiplyOptions.

EscapedGibbon: But if the options are the same as with multiply function, should I create a DivideOptions interface anyway? Or should I do it differently?

targos:
@stropitek Was it discussed to put these methods in the `compute` folder? That doesn't seem right to me as it doesn't contain similar methods.
I think `add` and `subtract` are similar.

src/compute/divide.ts
stropitek: It's fine to have some duplicated types.

If you don't want to duplicate types, you can always do something like this:
```ts
interface OperationOptionsBase {
  channels?: number[];
  out?: Image
}

interface  MultiplyOptions extends OperationOptionsBase {};
interface  DivideOptions extends OperationOptionsBase {};
```

targos: Yes, in general, 1 function with options = 1 interface

stropitek:
> Was it discussed to put these methods in the `compute` folder? That doesn't seem right to me as it doesn't contain similar methods.

No this wasn't discussed and I did not pay attention to that during the review.

> I think `add` and `subtract` are similar.

I agree

src/compare/multiply.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Channels where value will be multiplied.
+   * @default all channels
+   */
+  channels?: number[];
+  /**
+   * Image to which the resulting image has to be put.
+   */
+  out?: Image;
+}
```
stropitek, line 15: This is not good.

First the name is not good. Then you use the term "multiply" to document something that is used by multiply and divide.

There are many other methods that use those 2 options. To do this properly, the base should be shared with many other methods (so big refactoring).

So please just use 2 types `MultiplyOptions` and `DivideOptions`. It's ok that they have duplicate properties.

stropitek (changes requested):

stropitek (approved):

PR #462, opened 2024-05-17
feat: add bmp encoding

EscapedGibbon:


targos:
How did you create the images in `img/formats/bmp` ?

EscapedGibbon:
> How did you create the images in `img/formats/bmp` ?

The 5x5 I created with fast-bmp, the other two I took directly from fast-bmp repo.

src/save/encode.ts
```diff
@@ -29,13 +34,19 @@ const defaultPng: EncodeOptionsPng = { format: 'png' };
...
+  if (image instanceof Image) {
+    if (options.format === 'png') {
+      return encodePng(image, options.encoderOptions);
+    } else if (options.format === 'jpg' || options.format === 'jpeg') {
+      return encodeJpeg(image, options.encoderOptions);
+    } else {
+      throw new RangeError(`invalid format: ${options.format}`);
+    }
+  } else if (options.format === 'bmp') {
+    return encodeBmp(image);
```
stropitek, line 49: encodeBmp should accept both image and Mask (non-binary images can be stored in the bmp format), but should throw if an image is passed with a message that it is not implemented.

src/save/write.ts
```diff
@@ -115,12 +132,17 @@ export function writeSync(
...
+  options?: WriteOptionsBmp | WriteOptionsPng | WriteOptionsJpeg | WriteOptions,
 ): Uint8Array {
   if (!options || !('format' in options)) {
     const extension = nodePath.extname(destinationPath).slice(1).toLowerCase();
     if (extension === 'png' || extension === 'jpg' || extension === 'jpeg') {
+      if (image instanceof Mask) {
+        image = image.convertColor('GREY');
+      }
+      return encode(image, { ...options, format: extension });
+    } else if (extension === 'bmp' && image instanceof Mask) {
```
stropitek, line 145: Should not check for Mask here. Let the call to `encodeBmp` handle it (see my other comment)

src/save/encode.ts
```diff
@@ -29,13 +34,15 @@ const defaultPng: EncodeOptionsPng = { format: 'png' };
...
-  options: EncodeOptionsPng | EncodeOptionsJpeg = defaultPng,
+  image: Image | Mask,
+  options: EncodeOptionsBmp | EncodeOptionsPng | EncodeOptionsJpeg = defaultPng,
 ): Uint8Array {
   if (options.format === 'png') {
-    return encodePng(image, options.encoderOptions);
+    return encodePng(image as Image, options.encoderOptions);
   } else if (options.format === 'jpg' || options.format === 'jpeg') {
-    return encodeJpeg(image, options.encoderOptions);
+    return encodeJpeg(image as Image, options.encoderOptions);
```
EscapedGibbon, line 43: I am not sure if it's okay to do it this way. png and jpg do not support Mask, but I didn't find a throw in case a user tries to encode it with these formats. Plus, I believe using `as Image` is not a good practice, but I don't know how to do it otherwise.

targos: You added `Mask` to the input types of the `encode` function, so it's up to you to update the other encoders.
I think instead of throwing we should automatically convert the mask to something meaningful inside each encoder, like it's already done for jpeg when the image is not RGBA/8bit.

targos: The PNG format supports 1 bit per pixel but I don't know if the encoder does. If it doesn't, we can convert to grey/8bit.

EscapedGibbon: https://github.com/image-js/fast-png/blob/ff545ceb792a84ddd5366a0223fbe036e4a9c57d/src/PngEncoder.ts#L152-L159

Seems like encoder doesn't support 1 bit per pixel.

src/save/encodeBmp.ts
```diff
@@ -0,0 +1,40 @@
...
+      width: mask.width,
+      height: mask.height,
+      components: 1,
+      bitDepth: 1,
+      channels: 1,
+      data: compressedBitMask,
+    });
+  } else {
+    throw new TypeError('Image bmp encoding is not implemented.');
+  }
```
stropitek, line 39: can you move this condition to the top so that the main implementation does not have to be nested inside of an `if` statement

stropitek (approved):

Issue #464, opened 2024-05-22
Move automatic conversion from `write` to `encode`


targos:
I think each `encode***` function should handle its own automatic conversion based on what the encoder supports, or throw if the conversion cannot be done in a meaningful way.

stropitek:
It would be helpful to have a table in the documentation that shows how each color model is encoded

Have on column with the color model / bit depth of the original image and another with the color model / bit depth of `decode(encode(image))`

EscapedGibbon:
Here is what I came up with.

| DECODE | 8bits | 16bits | 32bits | Gray | RGB | Alpha |
|--------|-------|--------|--------|------|-----|-------|
| TIFF   | Yes   | Yes    | Yes    | Yes  | Yes | Yes   |
| PNG    | Yes   | Yes    | No     | Yes  | Yes | Yes   |
| JPEG   | Yes   | No     | No     | Yes  | Yes | No    |

BMP is not there because it is not decoded.TIFF and PNG can be decoded as anything. JPEG converts gray and rgb 8-bit images, but once decoded it sets colorModel as RGBA.

| ENCODE | 1bit | 8bit | 16bit | Gray | RGB | Alpha |
|--------|------|------|-------|------|-----|-------|
| PNG    | No   | Yes  | Yes   | Yes  | Yes | Yes   |
| JPEG   | No   | Yes  | No    | No   | Yes | No    |
| BMP    | Yes  | No   | No    | No   | No  | No    |

We don't support TIFF encoding. JPEG converts any image into 8 bit RGB. Again, important to note that it works only if the data's colorModel is set as RGBA. PNG has the same encoding capabilities as decoding. BMP can encode only binary images.

PR #466, opened 2024-05-24
refactor: move automatic conversion from write to encode

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/464

### DIFF
--- a/src/operations/__tests__/convertBitDepth.test.ts
+++ b/src/operations/__tests__/convertBitDepth.test.ts
@@ -65,3 +65,13 @@ test('throw if converting to same bit depth', () => {
     img.convertBitDepth(8);
   }).toThrow('cannot convert image to same bitDepth');
 });
+test('throw if converting to unsupported bit depth', () => {
+  const img = testUtils.createRgbaImage([
+    [256, 256, 256, 256, 512, 512, 512, 512],
+    [768, 768, 768, 768, 1024, 1024, 1024, 1024],
+  ]);
+
+  expect(() => {
+    img.convertBitDepth(1);
+  }).toThrow('This image bit depth is not supported: 1');
+});
--- a/src/operations/convertBitDepth.ts
+++ b/src/operations/convertBitDepth.ts
@@ -13,8 +13,12 @@ export function convertBitDepth(image: Image, newBitDepth: BitDepth): Image {
 
   if (newBitDepth === 16) {
     return convertToUint16(image);
-  } else {
+  } else if (newBitDepth === 8) {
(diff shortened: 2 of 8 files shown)


stropitek (commented):
Have you checked code coverage to see if there are non-covered lines in the encode functions?

EscapedGibbon:
> Have you checked code coverage to see if there are non-covered lines in the encode functions?

All the code is covered by tests. I removed the check on bit depth for png encoding, because once the mask changes color model, it automatically becomes 8 bit image.

EscapedGibbon:
I don't know, however, if there is a need for a case for 1 bit in convertBitDepth function. Technically, the user can try to change bit depth to 1. But nothing will happen, if you run `img = img.convertBitDepth(1)`. The image bit depth will not change. Should we add a throw to `convertBitDepth` if the image is not 8 or 16 bits?

stropitek:
> The image bit depth will not change. Should we add a throw to `convertBitDepth` if the image is not 8 or 16 bits?

I think in practice only `Mask` can have a bit depth of 1, not `Image`

stropitek (approved):

EscapedGibbon:
> > The image bit depth will not change. Should we add a throw to `convertBitDepth` if the image is not 8 or 16 bits?
> 
> I think in practice only `Mask` can have a bit depth of 1, not `Image`

I understand. But currently user can do something like this:
```ts
let image = new Image(4,4,{colorModel:"RGB"});
image = image.convertBitDepth(1);
//bit depth will remain 8
```
And there will be no warnings or throws. Normally, if a user wants a binary image, one will create a Mask, I get it.
 And such behavior doesn't break anything. But should we warn the user about it? Or is it an overkill?

targos:
Good point. I think we should throw an error

stropitek:
> I understand. But currently user can do something like this:
> 
> ```ts
> let image = new Image(4,4,{colorModel:"RGB"});
> image = image.convertBitDepth(1);
> //bit depth will remain 8
> ```

I didn't know this was possible. I agree with @targos, this should throw.

EscapedGibbon:
> > I understand. But currently user can do something like this:
> > ```ts
> > let image = new Image(4,4,{colorModel:"RGB"});
> > image = image.convertBitDepth(1);
> > //bit depth will remain 8
> > ```
> 
> I didn't know this was possible. I agree with @targos, this should throw.

Should I create separate issue or fix it in this PR?

stropitek:
> Should I create separate issue or fix it in this PR?

You can fix this here with a test.

Issue #470, opened 2024-06-07
Image.mean accept Point[] as option


lpatiny:
Improve Image.mean, Image.median, Image.variance accept an option `points`

PR #471, opened 2024-06-07
feat: add options to mean, median and variance 

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/470

### DIFF
--- a/src/compute/__tests__/mean.test.ts
+++ b/src/compute/__tests__/mean.test.ts
@@ -38,7 +38,7 @@ test('2x4 GREY image', () => {
 
   expect(result).toStrictEqual([1.5]);
 });
-test('compute mean from points array', () => {
+test('mean from points', () => {
   const image = testUtils.createGreyImage([
     [1, 2, 3, 0],
     [1, 2, 3, 0],
@@ -46,18 +46,13 @@ test('compute mean from points array', () => {
   const points = [
     { column: 0, row: 0 },
     { column: 1, row: 0 },
-    { column: 2, row: 0 },
-    { column: 3, row: 0 },
-    { column: 0, row: 1 },
-    { column: 1, row: 1 },
     { column: 2, row: 1 },
-    { column: 3, row: 1 },
   ];
   const result = image.mean({ points });
 
-  expect(result).toStrictEqual([1.5]);
+  expect(result).toStrictEqual([2]);
 });
-test('compute mean from points array in rgba image', () => {
+test('mean from points in rgba image', () => {
   const image = testUtils.createRgbaImage([
     [1, 2, 3, 0],
     [1, 2, 3, 0],
--- a/src/compute/__tests__/mean.test.ts
+++ b/src/compute/__tests__/mean.test.ts
@@ -38,3 +38,35 @@ test('2x4 GREY image', () => {
 
   expect(result).toStrictEqual([1.5]);
 });
+test('compute mean from points array', () => {
+  const image = testUtils.createGreyImage([
+    [1, 2, 3, 0],
+    [1, 2, 3, 0],
+  ]);
+  const points = [
+    { column: 0, row: 0 },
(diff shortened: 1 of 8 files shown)


stropitek (commented):
Variance will return the result of a 0-division if the array of points is empty. So `Infinity` or `-Infinity`. Did you think about what should happen in that case?

src/compute/mean.ts
```diff
@@ -1,19 +1,37 @@
...
+    for (let row = 0; row < image.height; row++) {
+      for (let column = 0; column < image.width; column++) {
+        for (let channel = 0; channel < image.channels; channel++) {
+          pixel[channel] += image.getValue(column, row, channel);
+        }
       }
     }
   }
-  return pixel.map((channel) => channel / image.size);
+  return pixel.map((channel) => channel / nbValues);
```
stropitek, line 36: Why are you calling your sum "channel"

stropitek:
> Variance will return the result of a 0-division if the array of points is empty. Did you think about what should happen in that case?

Same question is relevant for the other operations.

src/compute/mean.ts
EscapedGibbon: Isn't it what it is? We find a mean value of a pixel. Each value in this pixel is a channel. Although, calling it `channelSum` can be a better option.

EscapedGibbon:
> Variance will return the result of a 0-division if the array of points is empty. Did you think about what should happen in that case?

Indeed this should throw. Should we also check the validity of a coordinate? Right now the user can input any value there, regardless whether the point belongs to an image or not.

stropitek:
> Indeed this should throw.

I agree

> Should we also check the validity of a coordinate? Right now the user can input any value there, regardless whether the point belongs to an image or not.

I think it should throw as well. To be efficient, I suggest that you first convert the points to indices so that you don't do it twice, checking their value (compare with `size`) and then use `getValue` safely.

Please add test cases for each of those scenarios (no points, invalid points).

src/compute/__tests__/mean.test.ts
```diff
@@ -65,3 +66,27 @@ test('mean from points in rgba image', () => {
...
+    const result = image.mean({ points });
+    return result;
+  }).toThrow('Array of coordinates is empty.');
+});
+test("must throw if point's coordinates are invalid.", () => {
+  const image = testUtils.createRgbaImage([
+    [1, 2, 3, 0],
+    [1, 2, 3, 0],
+  ]);
+  const points: Point[] = [{ column: 0, row: 2 }];
```
stropitek, line 86: But if you pass `{column: 4, row:0}` which is also out of bound it does not throw right?

Can you add a test case +  change the implementation so that it also throws in that case?

src/compute/median.ts
```diff
@@ -19,12 +19,19 @@ export interface MedianOptions {
 export function median(image: Image, options?: MedianOptions): number[] {
   const pixel = new Array<number>(image.channels).fill(0);
   if (options) {
+    //console.log(options.points.length);
```
stropitek, line 22:
```suggestion
```

src/compute/__tests__/mean.test.ts
EscapedGibbon: > But if you pass `{column: 4, row:0}` which is also out of bound it does not throw right?
> 
> Can you add a test case + change the implementation so that it also throws in that case?

I think the way is to add a comparison between `point.column` and `image.width`. I am not sure I understand now what is the efficiency advantage of calculating the index in the first place. We can just use `point.column` and `point.row` to compare and then use either `getValue(point.column,point.row,channel)` or `getValueByPoint(point,channel)`.

stropitek: Yes there is no point anymore in calculating the index separately.

stropitek (approved):

Issue #469, opened 2024-06-07
Improve ROI - points, absolutePoints, relativePoints


lpatiny:
- [] - points becomes a Generator
- absolutePoints: use the Generator (points) in order to cache the array of absolute points. The result should be cached
- relativePoints: use the Generator (points) in order to cache the array of relative points. The result should be cached and will yield to the current result of points

PR #472, opened 2024-06-07
feat!: refactor points and add absolute coords calculation

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/469

### DIFF
--- a/src/roi/Roi.ts
+++ b/src/roi/Roi.ts
@@ -11,18 +11,6 @@ import { getMask, GetMaskOptions } from './getMask';
 import { getEllipse } from './properties/getEllipse';
 import { Border, Ellipse } from './roi.types';
 
-export interface PointsOptions {
-  /**
-   * Kind of coordinates that will be calculated and cached.
-   * For each kind the cache is different as well.
-   */
-  kind: 'relative' | 'absolute';
-}
-/**
- * Properties of borders of ROI.
- *
- */
-
 interface Computed {
   perimeter: number;
   borders: Border[]; // external and internal ids which are not equal to the current roi ID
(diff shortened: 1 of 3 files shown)


targos:
I didn't expect the generator to be a private method.
I thought we discussed:
- `.points()` generator function
- `.absolutePoints` getter with cache
- `.relativePoints` getter with cache

EscapedGibbon:
> I didn't expect the generator to be a private method. I thought we discussed:
> 
> * `.points(kind)` generator function
> * `.absolutePoints` getter with cache
> * `.relativePoints` getter with cache

We decided that it would be easier that the generator would treat both cases for absolute and relative coordinates. And since the code for both getters became similar, we decided to put it into one getter with a parameter that lets the user choose what kind of coordinates one wants. The coordinates are still cached separately though.

stropitek:
> We decided that it would be easier that the generator would treat both cases for absolute and relative coordinates.

That is not contradicting what @targos said. The point is that it should not be private.



> And since the code for both getters became similar, we decided to put it into one getter with a parameter that lets the user choose what kind of coordinates one wants.

It's no longer a "getter" if it takes parameters. Let's go with the initial idea: `relativePoints` and `absolutePoints` getters (otherwise we would have to find another name to differentiate the generator from the function returned the cached points.

src/roi/Roi.ts
```diff
@@ -292,25 +288,17 @@ export class Roi {
   }
   /**
    * Computes current ROI points.
-   * @returns Array of points. It's an array of tuples, each tuple being the x and y coordinates of the ROI point.
+   * @param kind - kind of coordinates.
+   * @returns Array of points with relative (coordinates are calculated based on ROI's point of origin) or absolute (coordinates are calculated based on the origin point of  an image/ROI map) ROI coordinates.
```
stropitek, line 292: This explains the difference between relative and coordinate and so should be explained at the level of the `kind` parameter.

Avoid making complicated sentences with a lot of content in parenthesis.

stropitek:
API SGTM but the explanations related to the new fields don't seem clear enough to me. We should discuss it IRL.

EscapedGibbon:
I wrote "Images/Masks" instead of "RoiMap" for absolute coordinates because RoiMap doesn't have `origin` property.

stropitek (approved):
Can you do the relative / absolute tests with the same data? It's nice to be able to compare the result of each with one another.

Issue #467, opened 2024-05-30
Multiple image analysis


lpatiny:
We have a sequence of images stored as a TIFF and would like to:
- make an average
- caclulate ROI on the average
- extract the average value of ROI on all the images

I'm wondering if this can be achieved using image-js-typescript.

EscapedGibbon:
This is too specific for a public API, with too many hidden things happening inside.

We will create a tutorial on how to extract the average value of ROI on all the images.

PR #473, opened 2024-06-14
feat: add computation of ROI mean values in stack

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/467


targos (changes requested):
This is too specific for a public API, with too many hidden things happening inside.

Maybe this should be a tutorial in the documentation?

EscapedGibbon:
> This is too specific for a public API, with too many hidden things happening inside.
> 
> Maybe this should be a tutorial in the documentation?

I think it is too specific as a standalone tutorial as well. We need also to add some documentation about how to work with Stack and what it is in general. But I think it can be a tutorial indeed. @stropitek @lpatiny  WDYT?

stropitek:
I agree it is way too specific to be implemented directyl into image-js.

I supppose you tested it on images with a real use case so it would be interesting to have a tutorial with the results in image-js-docs

PR #475, opened 2024-07-05
feat: add Stack class to public API

EscapedGibbon:
close: https://github.com/image-js/image-js-typescript/issues/474

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ export * from './operations';
 export * from './roi';
 export * from './save';
 export * from './Stack';
+export { decodeStack } from './stack/load/decodeStack';
 export * from './utils/utils.types';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -15,6 +15,8 @@ export * from './morphology';
 export * from './operations';
 export * from './roi';
 export * from './save';
+export * from './Stack';
+export * from './stack/index';
```
stropitek, line 19: For now, do not export all these methods and only export the `Stack` class.

It is confusing to have all those stack-specific function with generic names export at the first level.

Eventually we would want to import stack-related functions in a namespace: `import {xxx} from 'image-js/stack'`

stropitek (approved):

stropitek (approved):

PR #478, opened 2024-09-18
feat: add setBlendedPixel for draw* functions

EscapedGibbon:


src/Image.ts
```diff
@@ -437,7 +437,9 @@ export class Image {
     channel: number,
     value: number,
   ): void {
-    this.data[(row * this.width + column) * this.channels + channel] = value;
+    if (column >= 0 && column < this.width && row >= 0 && row < this.height) {
```
stropitek, line 440: We probably want to avoid adding checks here for performance reasons. This method is called very often, often in places where the call is already safe.

If we really need this, then I would rather do that in new method.

stropitek: I think the check actually doesn't do anything because assigning out of bounds on a Uint8Array or likewise is a noop.

CleanShot 2024-09-18 at 14 17 10

src/draw/drawCircleOnImage.ts
```diff
@@ -60,20 +61,22 @@ export function drawCircleOnImage(
...
   } else {
     if (radius === 1) {
-      newImage.setVisiblePixel(center.column, center.row, fill);
+      setBlendedVisiblePixel(newImage, center.column, center.row, {
+        color: fill,
+      });
     }
     circle(center.column, center.row, radius, (column: number, row: number) => {
-      newImage.setVisiblePixel(column, row, color);
+      setBlendedVisiblePixel(newImage, column, row, { color });
```
stropitek, line 79: You can create the color object outside of the callback to avoid recreating it for every pixel.

src/draw/drawPoints.ts
```diff
@@ -61,10 +62,11 @@ export function drawPoints(
   });
 
   for (const point of points) {
-    newImage.setVisiblePixel(
+    setBlendedVisiblePixel(
+      newImage,
       Math.round(origin.column + point.column),
       Math.round(origin.row + point.row),
-      color,
+      { color },
```
stropitek, line 69: ditto (I'm not adding other dittos for this, but there are more)

src/draw/drawRectangle.ts
```diff
@@ -105,8 +114,12 @@ export function drawRectangle(
         currentColumn < column + width - 1;
         currentColumn++
       ) {
-        newImage.setVisiblePixel(currentColumn, currentRow, fillColor);
-        newImage.setVisiblePixel(currentColumn, currentRow, fillColor);
+        /* setBlendedVisiblePixel(newImage, currentColumn, currentRow, {
+          color: fillColor,
+        });*/
```
stropitek, line 119:
```suggestion
```

src/utils/__tests__/setBlendedVisiblePixel.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  const image = testUtils.createGreyaImage([
+    [50, 255, 1, 200, 2, 3, 4, 200],
+    [20, 30, 5, 200, 6, 7, 8, 200],
+    [1, 2, 3, 200, 4, 5, 6, 200],
+  ]);
+  setBlendedVisiblePixel(image, 0, 5, { color: [40, 40, 40, 40] });
+  expect(image).toMatchImageData([
+    [50, 255, 1, 200, 2, 3, 4, 200],
+    [20, 30, 5, 200, 6, 7, 8, 200],
+    [1, 2, 3, 200, 4, 5, 6, 200],
```
stropitek, line 39: It would be easier to understand that what is expected is that the data does not change if you reused a variable for the calls to `createGreyaImage` and `toMatchImageData`

src/utils/setBlendedVisiblePixel.ts
stropitek, line 1: There seems to be a lot of repetition of logic that already is in `setBlendedPixel`

Why not just make the check on whether the pixel is visible or not, and if yes call `setBlendedPixel`?

src/draw/drawLineOnImage.ts
```diff
@@ -56,7 +57,7 @@ export function drawLineOnImage(
     Math.round(origin.column + to.column),
     Math.round(origin.row + to.row),
     (column: number, row: number) => {
-      newImage.setVisiblePixel(column, row, color);
+      setBlendedVisiblePixel(newImage, column, row, { color });
```
stropitek, line 60: ditto

src/utils/__tests__/setBlendedVisiblePixel.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  ]);
+});
+
+test('GREYA image: alpha different from 255', () => {
+  const image = testUtils.createGreyaImage([[50, 64]]);
+  setBlendedVisiblePixel(image, 0, 0, { color: [100, 128] });
+  const alpha = 128 + 64 * (1 - 128 / 255);
+  const component = (100 * 128 + 50 * 64 * (1 - 128 / 255)) / alpha;
```
stropitek, line 47: I wouldn't make a unit test that reapplies the calculus of the blending by hand like that. I don't see what it brings. If one wants to see how the calculus works, they can check the source code.

stropitek (approved):

Issue #476, opened 2024-07-30
image.toDataURL() is missing


lpatiny:
I could not find the way to convert an image toDataURL in the new library

Previous implementation:

https://github.com/image-js/image-js/blob/17d7e83096318901d998d2139d974a2237af137c/src/image/core/export.js#L121-L143

https://github.com/image-js/image-js/blob/legacy/src/image/core/__tests__/toDataURL.js

PR #481, opened 2025-02-25
feat: add base64 image encoding 

EscapedGibbon:
close: #476


EscapedGibbon:
CleanShot 2025-02-25 at 15 51 23

However, `Buffer.from()` cannot be used in browser and `TextDecoder` is unstable and doesn't work for Node18 and Node20.

EscapedGibbon:
> Can you compare with https://www.npmjs.com/package/@jsonjoy.com/base64 which claims to be compatible with node and browsers.

CleanShot 2025-02-28 at 08 51 32

 https://www.npmjs.com/package/@jsonjoy.com/base64  is much faster than current implementation. I used vitest testing tools and it seems it works in the browser as well. I also compared it with a function that @lpatiny sent me, which uses `uint8-base64` library and `TextDecoder`. Here's the code example for it:
```ts
function encodeWithUint8Encode(buffer:Uint8Array){
  const base64 = uint8Encode(buffer);
  const binString = new TextDecoder().decode(base64);
  return `data:image/png;base64,${binString}`;
}
```

targos:
Then let's use it!

src/save/encodeBase64.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { encode } from './encode.js';
+import type { ImageFormat } from './encode.js';
+/**
+ * Converts image into a base64 URL string.
+ * @param image - Image to get base64 encoding from.
+ * @param format - Image format.
+ * @returns base64 string.
+ */
+export function encodeBase64(image: Image, format: ImageFormat) {
```
targos, line 13:
```suggestion
export function encodeDataURL(image: Image, format: ImageFormat) {
```

And rename the file, tests, etc.

src/save/__tests__/encodeBase64.test.ts
```diff
@@ -0,0 +1,85 @@
+import { createRgbaImage } from '../../../test/testUtils.js';
```
targos, line 1:
```suggestion
```

`testUtils` are available globally

src/save/__tests__/encodeBase64.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  expect(base64Data).toMatchSnapshot();
+});
+
+test('legacy image-js test', () => {
+  const image = createRgbaImage([
+    [
+      255,
+      0,
+      0,
+      255, // red
```
targos, line 41: Can you use a template string variant? It will be more compact / readable. Example: https://github.com/image-js/image-js-typescript/blob/bd5cc06ea5e2e284a7531011d85efefc0635b8f9/test/__tests__/testUtils.test.ts#L129-L132

stropitek:
> also compared it with a function that @lpatiny sent me, which uses `uint8-base64` library and `TextDecoder`

Is that the implementation you used in the benchmark as "TextDecoder"?

EscapedGibbon:
> > also compared it with a function that @lpatiny sent me, which uses `uint8-base64` library and `TextDecoder`
> 
> Is that the implementation you used in the benchmark as "TextDecoder"?

Yes

stropitek:
Can you show the benchmark code you used?

EscapedGibbon:
> Can you show the benchmark code you used?

```ts

import { toBase64 } from "@jsonjoy.com/base64";
// eslint-disable-next-line import/no-extraneous-dependencies
import { encode as uint8Encode } from 'uint8-base64';
import {bench} from 'vitest';

import {readSync} from '../../load/read.js';
import {encode} from '../encode.js';


function testEncodeWithJoin(buffer:Uint8Array){
  const binaryArray = [];
  const format = 'png';
 for (const el of buffer) {
   binaryArray.push(String.fromCodePoint(el));
 }
 const binaryString = binaryArray.join('');
 const base64String = btoa(binaryString);
 const dataURL = `data:image/${format};base64,${base64String}`;
 return dataURL;
};

function encodeWithUint8Encode(buffer:Uint8Array){
  const base64 = uint8Encode(buffer);
  const binString = new TextDecoder().decode(base64);
  return `data:image/png;base64,${binString}`;
}

const buffer1 = encode(readSync("/Users/maxim/git/zakodium/qrcode-reader/dataset/testing/test.png"));
const buffer2 = encode(readSync("/Users/maxim/git/zakodium/qrcode-reader/dataset/testing/smallQR.png"));
const buffer3 = encode(readSync("/Users/maxim/git/zakodium/qrcode-reader/dataset/testing/doc2.png"));


if(typeof window === 'object'){
  console.log("Browser");
}else{
  console.log("Node");
}

console.assert(testEncodeWithJoin(buffer1) ===  `data:image/png;base64,${toBase64(buffer1)}` && encodeWithUint8Encode(buffer1) ===  testEncodeWithJoin(buffer1));

describe("benchmarking different functions buffer 1",()=>{
 bench("with join",()=>{
   testEncodeWithJoin(buffer1);
 })
  bench("with toBase64",() => {
  return `data:image/png;base64,${toBase64(buffer1)}`;
})
bench("with textDecoder",()=>{
  encodeWithUint8Encode(buffer1);
})

})

describe("benchmarking different functions buffer 2",()=>{
  bench("with join",()=>{
    testEncodeWithJoin(buffer2);
  })
   bench("with toBase64",()=>{
    return `data:image/png;base64,${toBase64(buffer2)}`;
 })
 bench("with textDecoder",()=>{
   encodeWithUint8Encode(buffer2);
 })
 })

 describe("benchmarking different functions buffer 3",()=>{
  bench("with join",()=>{
    testEncodeWithJoin(buffer3);
  })
   bench("with toBase64",() => {
   return `data:image/png;base64,${toBase64(buffer3)}`;
 })
 bench("with textDecoder",()=>{
   encodeWithUint8Encode(buffer3);
 })
 })
```

stropitek:
I don't think the benchmark you published is valid because you ran it in nodejs and the library will use `Buffer.toString()` in most cases when it detects it (won't use that only with very small data).

You should do `Buffer = undefined` before running the benchmark.

lpatiny:
You can see the Buffer here: https://github.com/jsonjoy-com/base64/blob/master/src/toBase64.ts#L11

Also take care to remove Buffer without importing the library (annoying because our rules reorder the code ...) like in:

```js
Buffer = undefined;

const { toBase64 } = require('@jsonjoy.com/base64');

const bytes = new Uint8Array(64 * 1024 * 1024).map((_, i) =>
  Math.floor(Math.random() * 256),
);

console.time('base64');
const base64 = toBase64(bytes);
console.timeEnd('base64');

console.log(base64.slice(0, 100));
```

targos:
> annoying because our rules reorder the code ...

When you use `import` syntax, it doesn't matter where the code is. The imports are always executed first.

lpatiny:
> > annoying because our rules reorder the code ...
> 
> When you use `import` syntax, it doesn't matter where the code is. The imports are always executed first.

Ok didn't know this. Thanks !

So the only way would be to 'await import' after removing Buffer ?

targos:
Or `import './someSetupFile.js'` or `node --import setup.js script.js` or other alternatives

EscapedGibbon:
Vitest has Buffer in jsdom environment because, apparently, there are dependencies that rely on it.
https://github.com/vitest-dev/vitest/blob/63e89702c1e4f61cf94227ba2795d513919c6b8f/packages/vitest/src/integrations/env/jsdom.ts#L73-L77

If I put the `Buffer = undefined` the tests will simply crash. Since we basically want to test this line here: 
https://github.com/jsonjoy-com/base64/blob/6bd3d6f063477680692c9c96fa7e0ebe407d3e6d/src/toBase64.ts#L7

Can I modify the checking in node modules' function JUST for this benchmarking, so that the `encodeSmall` gets called? I know for a fact that the imported package works in the browser, I just don't know how to remove `Buffer` from testing. Should I look for a different approach?

stropitek:
I don't have the context of why you used vitest for the benchmark.

Initially you used a library for benchmarking in Node.js which does not depend on jsdom.

EscapedGibbon:
Seems to be working. So, in a browser it seems like the fastest one is Luc's version of encoding.

CleanShot 2025-02-28 at 14 11 44

Here 's the code:

```ts
import  {toBase64}  from "@jsonjoy.com/base64";
import { encode as uint8encode } from "uint8-base64";
import assert from 'assert';
import {summary,bench,run} from 'mitata';
import { readSync,encode } from "image-js";

console.log(typeof Buffer);
function getDecodedData(id){
      // @ts-ignore
      const binaryString = atob(document.getElementById(id)?.src?.slice(22));
       const binaryData = new Uint8Array(binaryString.length);
       for(let element = 0; element < binaryString.length; element++){
        binaryData[element] = (binaryString.charCodeAt(element));
       }
       return binaryData;
     }

// Method 1: Using btoa + join
function encodeWithBtoaJoin(buffer){
 
  const binaryArray = new Array();
  for(let el of buffer){
    binaryArray.push( String.fromCharCode(el));
  }
   const binaryString = binaryArray.join('');
  const base64String = btoa(binaryString);
  const dataURL = `data:image/png;base64,${base64String}`;
  return dataURL;
}

// Method 2: Using uint8-base64 library
function encodeWithUint8Base64(buffer) {
  const base64 = uint8encode(buffer);
  const binString = new TextDecoder().decode(base64);
  return `data:image/png;base64,${binString}`;
}

// Method 3: Using @jsonjoy.com/base64 library
function encodeWithJsonJoyBase64(buffer) {
  return `data:image/png;base64,${toBase64(buffer)}`;
}

const data = getDecodedData('image2');

summary(() => {
  bench('btoa+join ', function* (ctx) {
    yield () => encodeWithBtoaJoin(data);
  });
 
  bench('uint8-base64', function* (ctx) {
    yield () => encodeWithUint8Base64(data);
  })
  
  bench('jsonjoy/base64 )', function* (ctx) {
    yield () => encodeWithJsonJoyBase64(data);
  })
 
});

await run();
```

targos:
What is the size of your image?

EscapedGibbon:
370x370 . Should i add a bigger image?

lpatiny:
> What is the size of your image?

It seems very small. I would use 10Mb for the tests. Based on my little benchmark (I added it on the [README](https://github.com/cheminfo/uint8-base64?tab=readme-ov-file#encoding-to-text-speed-test), 10Mb should take 10ms.

EscapedGibbon:
Ok, will do

EscapedGibbon:
CleanShot 2025-02-28 at 15 03 19

here are the results for 10mb array.

src/save/encodeDataURL.ts
```diff
@@ -0,0 +1,25 @@
...
+import { encode, defaultPng } from './encode.js';
+/**
+ * Converts image into Data URL string.
+ * @param image - Image to get base64 encoding from.
+ * @param options - Encoding options.
+ * @returns base64 string.
+ */
+export function encodeDataURL(
+  image: Image,
+  options: EncodeOptionsBmp | EncodeOptionsPng | EncodeOptionsJpeg = defaultPng,
```
EscapedGibbon, line 19: I kept the types the same way they were in `encode` function but should it be a single type that combines all encode<Type>Options interfaces?

package.json
```diff
@@ -43,6 +43,7 @@
   },
   "homepage": "https://github.com/image-js/image-js#readme",
   "dependencies": {
+
```
targos, line 46:
```suggestion
```

package.json
```diff
@@ -93,6 +94,7 @@
     "rimraf": "^6.0.1",
     "tailwindcss": "^3.4.17",
     "typescript": "~5.7.3",
+    "uint8-base64": "^0.1.1",
```
targos, line 97: You can update to v1.0.0: https://github.com/cheminfo/uint8-base64/releases/tag/v1.0.0

src/save/encode.ts
```diff
@@ -29,7 +29,7 @@ export interface EncodeOptionsJpeg {
 export interface EncodeOptionsBmp {
   format: 'bmp';
 }
-const defaultPng: EncodeOptionsPng = { format: 'png' };
+export const defaultPng: EncodeOptionsPng = { format: 'png' };
```
targos, line 32: You don't need to export this.

src/save/encodeDataURL.ts
targos, line 19:
```suggestion
  options?: EncodeOptionsBmp | EncodeOptionsPng | EncodeOptionsJpeg,
```

Just forward the options to `encode`, which manages the default value.

targos (approved):

PR #484, opened 2025-06-06
feat: implement perspective warp

EscapedGibbon:


stropitek:
The initial idea was to implement a function which returns a transform matrix, and to use that transform matrix to create the new image using `Image.transform` (https://image-js.github.io/image-js-typescript/classes/Image.html#transform)

The `transform` method already supports choosing the interpolation method.

EscapedGibbon:
There is a function in legacy version that calculates width and height with respect of image's aspect ratio. Is it okay to remove the function that calculates width and height to preserve ratio of the image and just specify desired width and height? openCV version doesn't have one, you need to put destination corner points there. We could do a function with 4 source points and 4 destination points.

EscapedGibbon:
Also the transform() function currently accepts 2x3 matrices for affine transformations. Should i modify the code so that it accepts only 3x3 matrices or it should accept both, and if input matrix is 2x3 i just add another row? After all, if i understand correcly, affine transformation is perspective transformation with [0,0,1] as last matrix row

src/geometry/getPerspectiveWarp.ts
```diff
@@ -0,0 +1,161 @@
...
+    throw new Error(
+      `The array pts must have four elements, which are the four corners. Currently, pts have ${pts.length} elements`,
+    );
+  }
+  const { width, height } = options;
+  const [tl, tr, br, bl] = order4Points(pts);
+
+  let widthRect;
+  let heightRect;
+  if (height && width) {
```
stropitek, line 47: If you think we should either specify both or none, then I think you should not let users specify just one but then override the value they provided. Throw an error or use the value they provided.

src/geometry/getPerspectiveWarp.ts
```diff
@@ -0,0 +1,166 @@
...
+    heightRect = height;
+  } else if (!height && !width) {
+    widthRect = Math.ceil(
+      Math.max(distance2Points(tl, tr), distance2Points(bl, br)),
+    );
+    heightRect = Math.ceil(
+      Math.max(distance2Points(tl, bl), distance2Points(tr, br)),
+    );
+  } else {
+    throw new Error(`${width ? 'Height' : 'Width'} is not defined.`);
```
stropitek, line 61: You can make a better error message:
- Be more explicit and complete. Explain the options' `width` and `height` properties are invalid with an explanation on how to use them correctly together.
- When using property names like `width` and `height` in an error message, I would advise to backquote them and respect the casing.

test/img/opencv/generate.py
```diff
@@ -91,3 +91,72 @@ def writeImg(name, img):
 kernelY = np.float32([[0.4, 0.5, 0.6, -0.3, -0.4]])
 dst = cv.sepFilter2D(img, ddepth=-1, kernelX=kernelX, kernelY=kernelY, borderType=cv.BORDER_REFLECT)
 writeImg('testConvolution.png', dst)
+
+
+# Image perspective warp
+
```
stropitek, line 97: Have you executed the script to regenerate the images?

EscapedGibbon: Yes, I executed it the way it was described in the README

stropitek (approved):

PR #489, opened 2025-07-25
test: fix encodeBmp equality checks

targos:


EscapedGibbon:
CleanShot 2025-07-25 at 09 42 15

@targos the cause of this is in the way we encode bmp but i wouldn't call it a bug. As you can see there is a difference in some of the fields between the image from data and test image. It happens because we usually store this data during bmp decoding. But, since image-js format doesn't store such specific data, encoder just fills it with default values.

Default values were chosen with GIMP specs in mind, so if we overwrite images in GIMP the tests will pass

CleanShot 2025-07-25 at 10 27 50

EscapedGibbon:
On the sidenote, I think it would be nice to include bmp decoder in image-js as well.

targos:
Thanks!

